PR #4, opened 2021-06-10
feat(windowservice): Made the windowService subjects private, added TSDoc and fix tests

IbenTesara:
When using the windowService in a project I noticed that I could just push a new value to the `width$`, whilst that property should only be updated when the window resizes, and I as a user should not be able to push anything on it.

I went through the code and saw that the BehaviorSubjects were public, so I made those private and returned the needed values as Observables. I also added an Observable for the current scroll position, which seemed useful to me when we for instance want to start loading specific content at specific scroll positions.

While I was at it, I added some TSDoc so that the doc in the readme would also be available in VSCode
afbeelding.

Finally I noticed the test weren't working so I fixed those as well.

### DIFF
--- a/projects/utils/src/lib/window-service/window.service.ts
+++ b/projects/utils/src/lib/window-service/window.service.ts
@@ -1,15 +1,35 @@
 import { DOCUMENT, isPlatformBrowser } from '@angular/common';
 import { Inject, Injectable, PLATFORM_ID } from '@angular/core';
-import { BehaviorSubject } from 'rxjs';
+import { BehaviorSubject, Observable } from 'rxjs';
 
 // @dynamic
 @Injectable({
 	providedIn: 'root',
 })
 export class WindowService {
-	public width$: BehaviorSubject<number> = new BehaviorSubject<number>(1200);
-	public scrollingUp$: BehaviorSubject<boolean> = new BehaviorSubject(true);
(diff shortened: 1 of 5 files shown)


projects/utils/src/lib/window-service/window.service.ts
```diff
@@ -1,15 +1,35 @@
...
-	public width$: BehaviorSubject<number> = new BehaviorSubject<number>(1200);
-	public scrollingUp$: BehaviorSubject<boolean> = new BehaviorSubject(true);
+	private widthSubject$: BehaviorSubject<number> = new BehaviorSubject<number>(1200);
+	private scrollingUpSubject$: BehaviorSubject<boolean> = new BehaviorSubject(true);
+	private currentScrollPositionSubject$: BehaviorSubject<number> = new BehaviorSubject<number>(0);
+
+	/**
+	 * Observable to get the window-width, defaults to 1200 when no window is defined
+	 */
+	public width$: Observable<number> = this.widthSubject$.asObservable();
```
DenisValcke, line 17: Aha good catch :ok_hand:

DenisValcke:
De comments zijn een goed idee, ik probeer dat ook zoveel mogelijk te doen maar eigenlijk moeten we daar wat meer consistent in zijn. Zeker in open-source packages en libraries.

DenisValcke (approved):

jsprds (approved):

PR #93, opened 2024-02-28
feat(ngx-table): Convert ngx-table to a standalone approach

IbenTesara:
In order to stay up to date with the current direction Angular is going in, all pipes, directives and components in the ngx-table package has been made to be standalone and to work with signals.

Given that effect and signal inputs aren't production ready yet; we decided not to include those yet


DenisValcke (approved):

projects/table/src/lib/table/ngx-table.component.ts
WHeirstrate, line 1: The order of the component modules is inconsistent.
1. `private` props
2. `private` hostbinding
3. `public` props
4. inputs
5. input `hostbinding`
6. inputs
7. outputs
8. `public` props

How about the following consistency: 2, 1, 3, 5, 4, 7?

WHeirstrate (changes requested):

WHeirstrate (approved):

PR #100, opened 2024-03-15
feat(forms): A MaxWordCount validator has been added

DenisValcke:
This validator will perform a simple check to see if a provided string has a min and/or max amount of words.

It trims spaces before and after the string using `String.trim()` and then counts using the `String.split()` and checking the length of the array it returns.


projects/forms/src/lib/validators/max-word-count/word-count.validator.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+		expect(WordCountValidator({ max: 3 })(new FormControl(' These Words Count '))).toBeNull();
+	});
+
+	it('should give an error when the value is a string that does exceed the maximum word count', () => {
+		expect(WordCountValidator({ max: 3 })(new FormControl('These Words Do Not'))).toEqual({
+			maxWordCountReached: { valid: false },
+		});
+	});
+
+	it('should not give an error when the value is a string that does not exceed the minimum word count', () => {
```
WHeirstrate, line 38: The title indicates **no** error should be returned. I believe this is a typo 😅

WHeirstrate (changes requested):

WHeirstrate (approved):

WHeirstrate (changes requested):
Any particular reason to include an `.nx/cache` folder? Perhaps this needs to be included in the `.gitignore`?

DenisValcke:
@WHeirstrate I have no idea why the `.nx` folder keeps popping up. I've removed it and added it to the `.gitignore` (see latest commit).

WHeirstrate (approved):

PR #108, opened 2024-03-26
fix(ngx-layout): Include the isActive state in the equal layout sizing

IbenTesara:
Fixes the equal sizing to also include the isActive state, which would result in incorrect views without it.

afbeelding
afbeelding


WHeirstrate (approved):
Looks good!

PR #109, opened 2024-03-27
fix(ngx-layout): Fix unnecessary emits in the ngx-configurable-layout

IbenTesara:
When the form gets patched after the initial writeValue, the form emitted unnecessarily.  This should be fixed.


WHeirstrate (approved):
Looks good!

PR #116, opened 2024-05-23
feat(ngx-cookies): Setup ngx-cookies

IbenTesara:
Adds a wrapper service to the CookieConsentV3 and provides the necessary observables to listen to the changes. It also provides a directive to handle this.

Tests will be added in a follow-up PR, as it's unclear how we're able to test this without essentially mocking everything and not really testing it (Maybe Cypress tests at one point?)


libs/cookies/src/lib/services/cookie/cookie.service.ts
```diff
@@ -0,0 +1,200 @@
...
+	 * Sets up the CookieConsent.
+	 *
+	 * For more information of on this configuration, check the documentation
+	 * https://cookieconsent.orestbida.com/essential/getting-started.html
+	 *
+	 * @param categories - The categories of cookies
+	 * @param language - The language settings of the cookies
+	 * @param configuration - An optional set configuration settings
+	 */
+	setupCookiesHandler(
```
DenisValcke, line 76: Kleintje, maar hier ben je een `public` vergeten. Moeten we misschien eens in de config opzetten dat het altijd moet 🤔

libs/cookies/src/lib/services/cookie/cookie.service.ts
```diff
@@ -0,0 +1,200 @@
...
+		configuration?: NgxCookieConfiguration
+	): void {
+		// Iben: If we're not in the browser, we early exit, so server-side rendering can be enabled
+		if (!isPlatformBrowser(this.platformId)) {
+			return;
+		}
+
+		// Iben: If no categories were provided, we early exit and throw an error.
+		if (!categories || Object.keys(categories).length === 0) {
+			console.error(
```
DenisValcke, line 88: Maakt dit iets kapot? Anders zou ik daar een warning van durven maken.

libs/cookies/src/lib/services/cookie/cookie.service.ts
```diff
@@ -0,0 +1,200 @@
...
+
+	/**
+	 * Whether a certain category of cookies was accepted
+	 *
+	 * @param name - The name of the category
+	 */
+	public hasAcceptedCategory(name: string): Observable<boolean> {
+		// Iben: Return every time the cookie consent has changed
+		return combineLatest([
+			this.cookiesConsented$.pipe(startWith(undefined)),
```
DenisValcke, line 146: Gewoon even inpikken, niet echt iets dat herwerkt moet worden. Ik vind dit interessant want ik heb altijd de neiging om impliciet `null` te gebruiken ipv. `undefined` omdat `undefined` gemanipuleerd kan worden.

libs/cookies/src/lib/directives/has-cookie/has-cookie.directive.ts
```diff
@@ -0,0 +1,175 @@
...
+
+		// Iben: Create a new onDestroyed handler
+		this.destroyed$ = new Subject();
+
+		// Iben: Render the views based on the correct state
+		combineLatest(
+			// Iben Check for each cookie if it is accepted
+			this.cookies.map((cookie) => {
+				// Iben: If no specific services were provided, we can just check the category
+				if (!cookie.services) {
```
DenisValcke, line 100: A lot of if/else nesting here. Is this something we can split up into smaller (functional) methods to make it more comprehensible?

libs/cookies/src/lib/directives/has-cookie/has-cookie.directive.ts
```diff
@@ -0,0 +1,175 @@
...
+			.pipe(
+				map((hasCookies) => {
+					return flatten(hasCookies).every((hasCookie) => hasCookie);
+				}),
+				tap((hasCookie) => {
+					// Iben: Clear the current view
+					this.viewContainer.clear();
+
+					// Iben: If there already is a component, destroy it so it can update correctly
+					if (this.componentRef) {
```
DenisValcke, line 121: A lot of if/else nesting here. Is this something we can split up into smaller (functional) methods to make it more comprehensible?

libs/cookies/src/lib/directives/has-cookie/has-cookie.directive.ts
```diff
@@ -0,0 +1,175 @@
...
+ */
+@Directive({
+	selector: '[hasCookie]',
+	standalone: true,
+})
+export class NgxHasCookieDirective {
+	/**
+	 * The destroyed state of the directive
+	 */
+	private destroyed$: Subject<void>;
```
DenisValcke, line 32: How do we feel about using our own packages within each other? Because we could use the DestroyComponent here 🤔

libs/cookies/README.md
```diff
@@ -0,0 +1,79 @@
+# NgxCookies
+
+This library provides a quick and easy wrapper for [CookieConsent V3](https://cookieconsent.orestbida.com).
```
DenisValcke, line 3: Maybe, when this package has been released in a 1.0.0 version, we should reach out to them to see if there is a possibility of them also mentioning us on their page?

DenisValcke:
> Tests will be added in a follow-up PR, as it's unclear how we're able to test this without essentially mocking everything and not really testing it (Maybe Cypress tests at one point?)

@IbenTesara I would follow a pragmatic approach when it comes to testing. The core of this library is the third party cookie-consent which should be tested in their library.

What I would do in our codebase is provide unit testing for small helper methods within the components/services and indeed Cypress testing would be a nice and welcome addition.

---

What I'm currently wondering is if we should make the test application server side rendered to see how this component behaves when SSR is enabled for an application. Since cookie-based functionality relies on the browser.

libs/cookies/src/lib/services/cookie/cookie.service.ts
IbenTesara: It does indeed, if you pass no categories the no popup gets shown, hence why.

libs/cookies/src/lib/directives/has-cookie/has-cookie.directive.ts
IbenTesara: It's something we have to do at one point; but we currently don't have a setup for the shared lib yet 😂 I didn't want to do it in this PR as it's already quite big, but I'll add a TODO comment!

libs/cookies/src/lib/directives/has-cookie/has-cookie.directive.ts
IbenTesara: Yeah that's doable, I'll look into it!

libs/cookies/README.md
IbenTesara: Good idea :D

libs/cookies/src/lib/tokens/has-cookies-fallback/has-cookies-fallback.ts
```diff
@@ -0,0 +1,8 @@
+import { InjectionToken, Type } from '@angular/core';
+
+import { NgxCookiesFallBackComponent } from '../../abstracts';
+
+/** A token to provide a fallback component that needs to be rendered when a specific (set of) cookie(s) was not accepted. */
```
WHeirstrate, line 5: This is oddly formatted. Any particular reason why?

WHeirstrate (changes requested):

libs/cookies/src/lib/directives/has-cookie/has-cookie.directive.ts
DenisValcke: That's fair. Let's keep this on the radar though? Maybe make a ticket on the project board?

libs/cookies/src/lib/directives/has-cookie/has-cookie.directive.ts
```diff
@@ -0,0 +1,175 @@
...
+	 * The else template in case the cookie is not accepted
+	 */
+	@Input() public set hasCookieElse(ngTemplate: TemplateRef<any>) {
+		this.elseTemplateRef = ngTemplate;
+		this.elseViewRef = null;
+		this.updateView();
+	}
+
+	constructor(
+		templateRef: TemplateRef<any>,
```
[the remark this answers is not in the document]
IbenTesara: Yes, this allows you to prevent anything outside the constructor from using it. 

It's been explained to me once why the templateRef was done this way in the original *ngIf implementation, but don't ask me why because I've forgotten 🙈

libs/cookies/src/lib/directives/has-cookie/has-cookie.directive.ts
DenisValcke: This is much better thanks!

libs/cookies/src/lib/services/cookie/cookie.service.ts
DenisValcke: Ok that's fair. Thanks for the info.

libs/cookies/src/lib/tokens/has-cookies-fallback/has-cookies-fallback.ts
IbenTesara: This is simply how you can do it for a single line, it's not better or worse than doing it over multiple lines really.

libs/cookies/src/lib/services/cookie/cookie.service.ts
IbenTesara: I updated the comment however, so that it's clearer to us why it throws an error 👍

DenisValcke (approved):
lgtm

WHeirstrate (approved):

PR #120, opened 2024-05-31
feat(ngx-utils): Add runInBrowser method to the WindowService

IbenTesara:


SergueySHD (approved):
https://tenor.com/btdon.gif

PR #121, opened 2024-06-04
feat(ngx-utils): Add NgxStorageService

IbenTesara:


apps/cookies-test/src/app/app.component.ts
```diff
@@ -63,6 +67,13 @@ export class AppComponent {
 	ngOnInit() {
 		this.ngxCookieService.setCookie({ name: 'authenticated', value: true });
 		this.ngxCookieService.setCookie({ name: 'kbo', value: true });
+
+		this.sessionService.storageEvents$.subscribe(console.log);
```
DenisValcke, line 71: Is the `cookies-test` application the right one for these examples? 🤔

DenisValcke (approved):
@IbenTesara I'll approve because functionaly this looks great, but can you take a look at my comment about the examples before merging?

apps/cookies-test/src/app/app.component.ts
IbenTesara: TBF, these apps aren't meant for examples, that's why they have a disclaimer at the top 😂 They're just used to test stuff for ourselves.

PR #137, opened 2024-06-10
upgrade(standalone-control-flow): Update the packages to work with standalone components and the new control flow

IbenTesara:


DenisValcke (approved):

apps/table-test/src/app/wrapper/wrapper.component.ts
```diff
@@ -1,13 +1,15 @@
-import { CommonModule } from '@angular/common';
 import { Component, Input } from '@angular/core';
 
 @Component({
 	selector: 'app-wrapper',
-	imports: [CommonModule],
+	imports: [],
```
WHeirstrate, line 5: It's in the test so it doesn't **have** to be fixed, but I'd remove this empty imports array.

WHeirstrate (approved):

PR #145, opened 2024-06-14
feat(ngx-layout): Add displayContent directive

IbenTesara:
Adds a directive that can provide a default fallback for loading, error and offline states.

Input on other possible states is always appreciated!

### DIFF
--- a/libs/layout/src/lib/abstracts/display-content/display-content.component.ts
+++ b/libs/layout/src/lib/abstracts/display-content/display-content.component.ts
@@ -1,0 +1,12 @@
+import { Directive, Input } from '@angular/core';
+
+/**
+ * An abstract class to use when implementing custom fallbacks for the NgxDisplayContentDirective
+ */
+@Directive()
+export abstract class NgxDisplayContentComponent {
+	/**
+	 * An optional data input to provide data to the fallback components. This property is typed as any.
+	 */
+	@Input() data: any;
+}
--- a/libs/layout/src/lib/directives/display-content/display-content.directive.ts
+++ b/libs/layout/src/lib/directives/display-content/display-content.directive.ts
@@ -1,0 +1,206 @@
+import {
+	AfterViewInit,
+	ChangeDetectorRef,
+	Directive,
+	Inject,
+	Input,
+	OnDestroy,
+	TemplateRef,
+	Type,
+	ViewContainerRef,
+} from '@angular/core';
+import { Subject, distinctUntilChanged, takeUntil, tap } from 'rxjs';
+import {
+	NgxDisplayContentConditions,
+	NgxDisplayContentConfiguration,
+	NgxDisplayContentOverrideConfiguration,
+	NgxDisplayContentStatus,
+} from '../../types';
+import { NgxDisplayContentConfigurationToken } from '../../tokens';
+import { NgxOnlineService } from '../../services';
+import { NgxDisplayContentComponent } from '../../abstracts';
+
+/**
+ * NgxDisplayContentDirective handles whether content should be shown based on a loading, error or offline condition by using provided default components or a provided override template
+ */
+@Directive({
+	selector: '[displayContent]',
+	standalone: true,
+})
+export class NgxDisplayContentDirective implements AfterViewInit, OnDestroy {
+	/**
+	 * A subject to handle the destroyed flow
+	 */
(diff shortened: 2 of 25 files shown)


WHeirstrate (approved):
LGTM

libs/layout/src/lib/directives/display-content/display-content.directive.ts
```diff
@@ -0,0 +1,199 @@
...
+		// Iben: Notify that the view needs to be updated
+		this.updateViewSubject.next();
+	}
+
+	constructor(
+		private readonly templateRef: TemplateRef<any>,
+		private readonly cdRef: ChangeDetectorRef,
+		private readonly viewContainer: ViewContainerRef,
+		private readonly onlineService: NgxOnlineService,
+		@Inject(NgxDisplayContentConfigurationToken)
```
DenisValcke, line 85: Just aesthetics but I would move this one to the top or bottom of the list.

libs/layout/src/lib/directives/display-content/display-content.directive.ts
```diff
@@ -0,0 +1,199 @@
...
+	/**
+	 * Holds the conditions we use to render the component
+	 */
+	private conditions: NgxDisplayContentConditions = {};
+
+	/**
+	 * Renders the item or a default fallback based on the provided conditions
+	 */
+	@Input() set displayContent(conditions: NgxDisplayContentConditions) {
+		// Iben: Update the conditions, but use a merge so that the online status can be preserved
```
DenisValcke, line 55: I would move this to a private method and re-use it:
```typescript
// Iben: Update the conditions, but use a merge so that the online status can be preserved
if (conditions) {
	this.conditions = {
		...this.conditions,
		...conditions,
	};
}

// Iben: Notify that the view needs to be updated
this.updateViewSubject.next();
```

libs/layout/src/lib/directives/display-content/display-content.directive.ts
```diff
@@ -0,0 +1,199 @@
...
+		private readonly configuration: NgxDisplayContentConfiguration
+	) {
+		// Iben: If we want to listen to the online status, we set up a listener to the status of the application
+		if (configuration.listenToOnlineStatus) {
+			this.onlineService.online$
+				.pipe(
+					distinctUntilChanged(),
+					tap((online) => {
+						// Iben: Update the conditions with the new status
+						this.conditions = {
```
DenisValcke, line 95: private method can be use here as well:
```typescript
this.setConditions({ offline: !online });
```

libs/layout/src/lib/abstracts/display-content/display-content.component.ts
```diff
@@ -0,0 +1,12 @@
...
+
+/**
+ * An abstract class to use when implementing custom fallbacks for the NgxDisplayContentDirective
+ */
+@Directive()
+export abstract class NgxDisplayContentComponent {
+	/**
+	 * An optional data input to provide data to the fallback components
+	 */
+	@Input() data: any;
```
DenisValcke, line 11: I would change this to a `never` or `unknown`, or make a mention in the comments that this is an intended `any`.

libs/layout/src/lib/services/online-service/online.service.ts
```diff
@@ -0,0 +1,54 @@
...
+	/**
+	 * An observable that emits whenever the application is on or offline
+	 */
+	public readonly online$: Observable<boolean> = this.onlineSubject.asObservable();
+
+	constructor(@Inject(PLATFORM_ID) platformId: string) {
+		// Iben: When we're in the browser, listen to the online and offline status of the application
+		if (isPlatformBrowser(platformId)) {
+			// Iben: Handle the on and offline status of the application
+			fromEvent(window, 'online')
```
DenisValcke, line 29: Not a fan of using the `window` directly here. Maybe create an issue that, when we are able to, we use the `WindowService` here to avoid potential problems.

For now it's okay. Just a to do to keep in mind.

DenisValcke (approved):

libs/layout/src/lib/services/online-service/online.service.ts
IbenTesara: Yeah, completely agree; it does use the same logic as the `WindowService` does, so this should work.

PR #152, opened 2024-06-24
feat(ngx-layout): Add tour service

IbenTesara:
https://imgur.com/a/oc0OLoA

Adds a light-weight easy to customize help tour lib


libs/tour/src/lib/services/tour-service/tour.service.ts
```diff
@@ -0,0 +1,486 @@
...
+	 * Whether the tour has ended as a subject
+	 */
+	private readonly tourEndedSubject: Subject<void> = new Subject();
+
+	/**
+	 * The currently active tour as a subject
+	 */
+	private readonly currentTourSubject: BehaviorSubject<NgxTourStep[]> = new BehaviorSubject<
+		NgxTourStep[]
+	>(undefined);
```
DenisValcke, line 100: Might be better not to use an implicit `undefined` but use `null` instead?
(also applies to other code in the PR)

libs/tour/src/lib/services/tour-service/tour.service.ts
```diff
@@ -0,0 +1,486 @@
...
+		tour: NgxTourStep[],
+		onClose?: NgxTourAction,
+		startIndex = 0
+	): Observable<NgxTourInteraction> {
+		// Iben: Save the current scroll position so we can return to it when we close the tour
+		this.runInBrowser(() => {
+			this.startingScrollPosition = window.scrollY;
+		});
+
+		// Iben: Loop over the tour and set subjects for all elements that haven't been found yet
```
DenisValcke, line 167: Maybe switch to `reduce` to create a new array each time and get rid of the reference.
```typescript
  // Iben: Loop over the tour and set subjects for all elements that haven't been found yet
  this.elements = tour.reduce((acc, step) => {
    if (!this.elements[step.tourItem]) {
      return {
        ...acc,
       [step.tourItem] = new BehaviorSubject<NgxTourItemDirective>(undefined);
      };

      return acc;
    }
  });
```

libs/tour/src/lib/services/tour-service/tour.service.ts
```diff
@@ -0,0 +1,486 @@
...
+
+		// Iben: Create the box around the item
+		const box = `${positionLeft}px ${positionTop}px, ${positionRight}px ${positionTop}px, ${positionRight}px ${positionBottom}px, ${positionLeft}px ${positionBottom}px, ${positionLeft}px ${positionTop}px`;
+
+		// Iben: Clip the box out of the backdrop
+		backdrop.style.clipPath = `polygon(evenodd, 0% 0%, 0% 100%, 100% 100%, 100% 0%, 0% 0%,${box})`;
+	}
+
+	//TODO: Iben: Remove this function in service of the WindowService once it is shared
+	private runInBrowser(callBack: () => void) {
```
DenisValcke, line 481: return type

libs/tour/src/lib/services/tour-service/tour.service.ts
```diff
@@ -0,0 +1,486 @@
...
+	}
+
+	/**
+	 * Registers a template element so it can be highlighted
+	 *
+	 * @param element - The highlighted element
+	 */
+	public registerElement(element: NgxTourItemDirective): void {
+		// Iben: Check if there is already an item with the current id. If it is, update it, if not, make a new BehaviorSubject
+		const elementSubject = this.elements[element.tourItem];
```
DenisValcke, line 210: Can `element` be undefined or not an object?

DenisValcke (approved):

libs/tour/src/lib/services/tour-service/tour.service.ts
IbenTesara: After internal discussions we decided to keep this as is.

libs/tour/src/lib/services/tour-service/tour.service.ts
IbenTesara: I left these subjects as is, as these can later on be changed by other values; but in the other places I updated it to be null 👍

PR #156, opened 2024-06-25
fix(ngx-tour): Fix issue with inconsistent scroll behaviour and prevent issues with missing the currentStep

IbenTesara:
- Prevents the tour from crashing if there for some reason isn't a current step
- Always scroll back to the top before each step so that when we move back and forth during the tour, the look and feel is consistent.

### DIFF
--- a/libs/tour/src/lib/services/tour-service/tour.service.ts
+++ b/libs/tour/src/lib/services/tour-service/tour.service.ts
@@ -333,6 +333,11 @@ export class NgxTourService implements OnDestroy {
 			this.overlayRef = undefined;
 		}
 
+		// Iben: Early exit if there's no current step
+		if (!currentStep) {
+			return of('close');
+		}
+
 		// Iben: Get the previous step, if it exists
 		const previousStep = this.currentStepSubject.getValue();
 		const previousItem = this.elements[previousStep?.tourItem]?.getValue();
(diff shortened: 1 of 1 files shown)


libs/tour/src/lib/services/tour-service/tour.service.ts
```diff
@@ -417,8 +422,11 @@ export class NgxTourService implements OnDestroy {
 		this.previousStepSubject.next(this.currentStepSubject.getValue());
 		this.currentStepSubject.next(currentStep);
 
-		// Iben: Scroll item into view if it's not in view
 		this.runInBrowser(() => {
+			// Iben: Scroll to the top before each step to get consistent behavior when going back and forth
+			window.scrollTo({ top: 0 });
```
WHeirstrate, line 427: Why is this needed? Which issue does this solve? If stap 1 and 2 are both outside the current viewport, the user will

1. Scroll down into the step
2. Scroll to the top
3. Scroll into the second step

They will be rubber banded with each step

IbenTesara: Given that the scroll to top goes instantly, you do not get a rubber-banded feeling at al. It wouldn't have been added if it gave that visual effect 😅 

As mentioned in the comment, if you don't do this, you get an inconsistent behaviour. Here's an example.

1. At the start, the item is in view. We're not scrolling to the element, as we don't have to.
afbeelding

2. We continue through the tour, and navigate to an element that's lower, so we scroll to that element. Now the above element is out of the viewport. Which isn't an issue.

3. Instead of continuing now, we however go back. But now, the item is not in the viewport and we scroll to it. The browser considers it to be in view, but now you get a completely different view than you had initially.
afbeelding

If you don't scroll back to the top before each step, you get inconsistent views between going back and forth, which is not something we want. We want the effect to be the same regardless of whether we're going back or forward.

WHeirstrate (approved):

libs/tour/src/lib/services/tour-service/tour.service.ts
WHeirstrate: As long as the user does not notice, I am a-ok with this implementation! 🙂

PR #165, opened 2024-07-01
feat(dispatch-data-to-store): add function overloads and allow SSOT by using selector

WHeirstrate:
## Explanation
The SSOT principle was not properly respected. The `dispatchDataToStore` util will not return anything in and of itself, a selector has to be provided. If not provided, the method will return `Observable<null>`.

## Examples

The function overloads make sure that something like this will not be allowed:
```Typescript
dispatchDataToStore(actions.amount, undefined, of([]), this.store)
```
The error will indicate that no contract of any overload matches. The `amount` is of type number, which indicates a `BaseStoreAsset<number>`, while the data provided is of type `Array`. The overloads will prevent this from passing, whereas previously this would fail to set the data in the store.

The error looks like this:
Screenshot 2024-07-01 at 08 21 22

The function overloads also make sure a `BaseStoreAsset` cannot have `update` or `add` as dispatchType, for they are only available on `EntityStoreAssets`.

## Possible improvements
Moving the function overloads to a type definition file `dispatch-data-to-store.util.d.ts`, for example.

### DIFF
--- a/libs/store/src/lib/store/utils/dispatch-data-to-store/dispatch-data-to-store.util.ts
+++ b/libs/store/src/lib/store/utils/dispatch-data-to-store/dispatch-data-to-store.util.ts
@@ -1,21 +1,93 @@
 import { Store } from '@ngrx/store';
-import { Observable, throwError } from 'rxjs';
-import { catchError, finalize, tap } from 'rxjs/operators';
+import { Observable, of, throwError } from 'rxjs';
+import { catchError, finalize, switchMap, tap } from 'rxjs/operators';
+import {
+	BaseStoreActions,
+	BaseStoreSelectors,
+	EntityStoreActions,
+	EntityStoreSelectors,
+} from '../../interfaces';
+
+// -------------------
+// Overloads
+// -------------------
+// BASE STORE
+export function dispatchDataToStore<DataType = any>(
+	dispatchAction: BaseStoreActions<DataType>,
+	selectAction: BaseStoreSelectors<DataType>,
+	data: Observable<DataType>,
+	store: Store,
+	dispatchType: 'set'
+): Observable<DataType>;
+export function dispatchDataToStore<DataType = any>(
+	dispatchAction: BaseStoreActions<DataType>,
+	selectAction: BaseStoreSelectors<DataType>,
+	data: Observable<DataType>,
+	store: Store
+): Observable<DataType>;
+// NO SELECTOR
+export function dispatchDataToStore<DataType = any>(
+	dispatchAction: BaseStoreActions<DataType>,
+	selectAction: null | undefined,
+	data: Observable<DataType>,
+	store: Store,
+	dispatchType: 'set'
+): Observable<null>;
+export function dispatchDataToStore<DataType = any>(
+	dispatchAction: BaseStoreActions<DataType>,
+	selectAction: null | undefined,
+	data: Observable<DataType>,
+	store: Store
+): Observable<null>;
+// ENTITY STORE
+export function dispatchDataToStore<DataType = any>(
+	dispatchAction: EntityStoreActions<DataType>,
+	selectAction: EntityStoreSelectors<DataType>,
+	data: Observable<DataType[]>,
+	store: Store,
+	dispatchType: 'set' | 'add' | 'update'
+): Observable<DataType[]>;
+export function dispatchDataToStore<DataType = any>(
+	dispatchAction: EntityStoreActions<DataType>,
+	selectAction: EntityStoreSelectors<DataType>,
+	data: Observable<DataType[]>,
+	store: Store
+): Observable<DataType[]>;
+// NO SELECTOR
+export function dispatchDataToStore<DataType = any>(
+	dispatchAction: EntityStoreActions<DataType>,
+	selectAction: null | undefined,
+	data: Observable<DataType[]>,
+	store: Store,
+	dispatchType: 'set' | 'add' | 'update'
+): Observable<null>;
+export function dispatchDataToStore<DataType = any>(
+	dispatchAction: EntityStoreActions<DataType>,
+	selectAction: null | undefined,
+	data: Observable<DataType[]>,
+	store: Store
+): Observable<null>;
+
+// -------------------
+// Definition
+// -------------------
 /**
  * Dispatches data to the store based on a provided action and Observable. Loading and error state will be handled by default.
  *
  * @param dispatchAction - The action we wish to use to dispatch data to the store
+ * @param selectAction - The selector we wish to use to select the data from the store. If not provided, `null` will be returned.
  * @param data - The data we wish to dispatch to the store
  * @param store - The store we wish to dispatch the data to
  * @param dispatchType - Whether we wish to set, add or update the data. By default this is `set`
- * @return {*}  {Observable<DataType>}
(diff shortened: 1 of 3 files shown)


libs/store/src/lib/store/utils/dispatch-data-to-store/dispatch-data-to-store.util.ts
```diff
@@ -1,37 +1,159 @@
...
 		tap((payload) => {
-			if (dispatchType === 'set') {
-				store.dispatch(dispatchAction.set({ payload }));
-			} else if (dispatchType === 'add') {
-				store.dispatch(dispatchAction.add({ payload }));
-			} else {
-				store.dispatch(dispatchAction.update({ payload }));
+			switch (dispatchType) {
+				case 'set': {
+					return payload instanceof Array
```
IbenTesara, line 100: You're making too big of an assumption here. Yes, `ngrx` strongly suggests using the entity store for arrays, but it's not mandatory. So people are still allowed to use the regular approach to use arrays. This package facilitates the process of `ngrx`, so we cannot prevent people from using it like that.

These checks will have to go, because they're too strongly set.

libs/store/src/lib/store/utils/dispatch-data-to-store/dispatch-data-to-store.util.ts
```diff
@@ -1,37 +1,159 @@
...
+								})
+						  )
+						: store.dispatch(
+								(dispatchAction as BaseStoreActions<DataType>).set({
+									payload,
+								})
+						  );
+				}
+				case 'add': {
+					if (!(payload instanceof Array)) {
```
IbenTesara, line 113: This check is also not really correct. If you would have an array of arrays in your store, you can no longer use the add method.

libs/store/src/lib/store/utils/dispatch-data-to-store/dispatch-data-to-store.util.ts
```diff
@@ -1,37 +1,159 @@
...
+			}
+		}),
+		switchMap((payload) => {
+			if (!selectAction) {
+				return of(null);
 			}
+
+			// Wouter: Get the value recently set to the store. This is done to keep the Store as SSOT.
+			return store.select(
+				payload instanceof Array
```
IbenTesara, line 151: I think it's better to run a check on the selectAction rather than on the payload, because again, that's not possible with how the store works.

You'll be better off doing something like `selectAction['selectAll'] ? selectAction.selectAll : selectAction.select`

libs/store/src/lib/store/utils/dispatch-data-to-store/dispatch-data-to-store.util.ts
```diff
@@ -1,37 +1,159 @@
...
+					? (selectAction as EntityStoreSelectors).selectAll
+					: (selectAction as BaseStoreSelectors).select
+			);
 		}),
-		// Iben: Catch the error and dispatch it to the store
+		// Iben: Catch the error and dispatch its to the store
```
IbenTesara, line 156: Why `its`? 😅 You want to dispatch the error to the store, aka `it`

libs/store/src/lib/store/utils/dispatch-data-to-store/dispatch-data-to-store.util.ts
```diff
@@ -1,37 +1,159 @@
...
+								)
+						);
+					}
+
+					return store.dispatch(
+						(dispatchAction as EntityStoreActions<DataType>).add({ payload })
+					);
+				}
+				case 'update': {
+					if (payload instanceof Array) {
```
IbenTesara, line 127: Same for this check

libs/store/src/lib/store/utils/dispatch-data-to-store/dispatch-data-to-store.util.ts
```diff
@@ -1,37 +1,159 @@
...
 	// Iben: Set, add or update the data according to the provided dispatch type
 	return data.pipe(
 		tap((payload) => {
-			if (dispatchType === 'set') {
-				store.dispatch(dispatchAction.set({ payload }));
-			} else if (dispatchType === 'add') {
-				store.dispatch(dispatchAction.add({ payload }));
-			} else {
-				store.dispatch(dispatchAction.update({ payload }));
+			switch (dispatchType) {
```
IbenTesara, line 98: Given that a switch case or an if else structure in this case makes no difference, I'd keep it the way it is if it isn't necessary to change it.

Changing these kinds of things in a PR without real cause makes it harder to review PRs, because you have to first try to figure out why it was done in relation to the changes, and then realize it wasn't really needed to do so.

libs/store/src/lib/store/utils/dispatch-data-to-store/dispatch-data-to-store.util.ts
```diff
@@ -1,37 +1,159 @@
...
+export function dispatchDataToStore<DataType = any>(
+	dispatchAction: BaseStoreActions<DataType>,
+	selectAction: null | undefined,
+	data: Observable<DataType>,
+	store: Store,
+	dispatchType: 'set'
+): Observable<null>;
+export function dispatchDataToStore<DataType = any>(
+	dispatchAction: BaseStoreActions<DataType>,
+	selectAction: null | undefined,
```
IbenTesara, line 38: I'm not that familiar with these overloads, but what will be the effect of putting the selectAction here on the existing implementation?

It's one thing that the result of this PR forces you to change the return types, but completely changing the order of this implementation forcing you to pass undefined as a second parameter would make it a lot harder to work wit.

I'd personally expect the overload to be either with the dispatchAction and one without it?

IbenTesara (changes requested):

libs/store/src/lib/store/utils/dispatch-data-to-store/dispatch-data-to-store.util.ts
WHeirstrate: Of course, no idea how that got changed! 😅

libs/store/src/lib/store/utils/dispatch-data-to-store/dispatch-data-to-store.util.ts
WHeirstrate: Yes, that would be possible, I presume! 👍

libs/store/src/lib/store/utils/dispatch-data-to-store/dispatch-data-to-store.util.ts
IbenTesara: True, I don't remember what I was referring to tbh 😅

IbenTesara:
To address the comments regarding the errors that are thrown here in the dispatch data to store. I've checked but basically came to the conclusion that all of these errors that are thrown are unnecessary.

1. The typing should block it first and foremost. If that is not the case yet, then we should ensure that the typing prevents people from passing the wrong data before building the application.
2. If the wrong kind of data is passed because the typing was not done correctly, NGRX simply does not update the state and throws their own errors where needed. We don't need to add additional errors here, NGRX already has us covered.
3. The array check on the add action clashes with the current implementation in the entity-adapter.util

In summary, these errors are not needed and if we want to improve the behaviour here, it should be through typing first and foremost.

PR #180, opened 2024-07-12
feat(ngx-layout): provide a service that allows subscribing to one or more media query

WHeirstrate:
# Description

This PR introduces a new service. It allows the user to provide one or more media queries whose changes will be emitted onto a stream. This stream can be subscribed to, but requires an id (the same that was used upon setting the query). If the id of the triggered breakpoint, the id of the triggered query will be emitted onto the stream. Those subscribers whose id matches the id that was pushed onto the stream, will be triggered and will receive the new query value (a boolean).

# Demo

https://github.com/user-attachments/assets/d492f1ec-f169-413d-b5ac-eb7676ec120d

### DIFF
--- a/libs/utils/README.md
+++ b/libs/utils/README.md
@@ -181,6 +181,12 @@ This service provides a SSR proof Observable based approach to both the local- a
 
 [Full documentation.](src/lib/services/storage-service/storage.service.md)
 
+#### NgxMediaQueryService
+
+This service provides a SSR proof way to set and subscribe to the window's media query changes.
+
+[Full documentation.](src/lib/services/media-query/query.service.md)
+
 ### Abstracts
 
 #### NgxQueryParamFormSyncComponent
--- a/README.md
+++ b/README.md
@@ -57,13 +57,19 @@ This repo contains general usage libraries for shd Angular projects. Those libra
     -   This library provides a quick and easy template based table builder using the Angular CDK Table.
 
 -   i18n (`@studiohyperdrive/ngx-i18n`):
+
     -   This library provides a lazy-loaded modular approach to translations.
 
 -   cookies (`@studiohyperdrive/ngx-cookies`):
+
     -   This library provides a quick and easy wrapper for [CookieConsent V3](https://cookieconsent.orestbida.com).
 
 -   layout (`@studiohyperdrive/ngx-layout`):
-    -   `ngx-layout` is a collection of Angular components related to layout.
+
+    -   This library is a collection of Angular components related to layout.
+
+-   tour (`@studiohyperdrive/ngx-tour`):
+    -   This library provides the tools to build a guided walkthrough of one or more pages.
 
 You can find detailed explanations in their respective README’s.
 
@@ -72,7 +78,7 @@ You can find detailed explanations in their respective README’s.
 ### Add a new library
 
 To add a new library, consult the Angular CLI documentation:
-https://angular.io/guide/creating-libraries
+https://angular.dev/tools/libraries/creating-libraries
 
 After adding your library, make sure to check the karma set-up is consistent with other projects.  
 Check the following files:
--- a/apps/layout-test/src/app/app.component.ts
+++ b/apps/layout-test/src/app/app.component.ts
@@ -1,6 +1,7 @@
 import { Component } from '@angular/core';
 
 import { RouterModule } from '@angular/router';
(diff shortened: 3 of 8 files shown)


libs/layout/src/lib/services/media-query/mediaquery.service.ts
```diff
@@ -0,0 +1,103 @@
...
+	constructor(private readonly windowService: WindowService) {}
+
+	/**
+	 * Register a list of media queries that need to be tracked by the service.
+	 *
+	 * @param queries - A list of media queries that should be registered with the service.
+	 */
+	public registerMediaQueries(...queries: [id: string, query: string][]): void {
+		this.windowService.runInBrowser(({ browserWindow }) => {
+			for (let [id, query] of queries) {
```
WHeirstrate, line 41: @IbenTesara remarked:
> I would add an early exit when a listener has already been registered on a certain `id`. Currently, the listener of a certain `id` will be overwritten without removing the previous one. This could lead to memory leaks.
> Example: setting two queries with the same id will result in the latter being destroyed in the onDestroy, while the first one will continue to exist and cause a memory leak.

libs/layout/src/lib/services/media-query/mediaquery.service.ts
```diff
@@ -0,0 +1,103 @@
...
+import { WindowService } from '@ngx/utils';
+import { filter, map, Observable, ReplaySubject } from 'rxjs';
+
+/**
+ * A service that can be used to track media queries and their changes. It exposes a method
+ * to register media queries, which takes an array of tuples with the id of the media query
+ * and the query itself. The service will then emit the id of the media query that has
+ * changed when subscribed to the `getMatchingQuery$` method.
+ */
+@Injectable()
```
WHeirstrate, line 11: @IbenTesara remarked:
> If we'd be supporting multiple matchingQueries, perhaps it would be more sensible to define the queries (and listeners) globally (in the `app.component.ts`), instead of setting them in each component that requires their use.

libs/layout/src/lib/services/media-query/mediaquery.service.ts
WHeirstrate: Good catch!

libs/layout/src/lib/services/media-query/mediaquery.service.ts
```diff
@@ -0,0 +1,103 @@
+import { Injectable, OnDestroy } from '@angular/core';
+import { WindowService } from '@ngx/utils';
```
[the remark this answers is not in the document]
IbenTesara: Alternatively, move this to the utils.

libs/layout/src/lib/services/media-query/mediaquery.service.ts
```diff
@@ -0,0 +1,103 @@
...
+		});
+	}
+
+	/**
+	 * Pass the id of the query whose changes need to be listened to.
+	 *
+	 * @param id - The id of the media query that should be checked.
+	 * @returns An observable that emits a boolean value whenever the requested media query changes.
+	 */
+	public getMatchingQuery$(id: string): Observable<boolean> {
```
IbenTesara, line 75: I would throw an error here if no query with the id was registered.

IbenTesara (changes requested):

libs/layout/README.md
IbenTesara, line 1: Maybe just revert the changes here, they're not necessary and just bloat the PR.

libs/utils/README.md
```diff
@@ -14,9 +14,9 @@ npm install @studiohyperdrive/ngx-utils
 
 This package will follow a semver-like format, `major.minor.patch`, in which:
 
-- `major`: Follows the Angular major version
-- `minor`: Introduces new features and (potential) breaking changes
-- `patch`: Introduces bugfixes and minor non-breaking changes
+-   `major`: Follows the Angular major version
```
IbenTesara, line 17: Is this really necessary? Because now we get an inconsistency between the readmes 🙈 If it's needed it's better to just do them all at once in a separate PR.

IbenTesara (approved):
PR looks good, just two small comments to take care of before publishing. I'll approve it already though. 👍

libs/utils/README.md
WHeirstrate: It is not, this was auto changed. I'll revert

PR #181, opened 2024-07-15
build(ngx-tools): v18.1.0

WHeirstrate:


libs/utils/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@studiohyperdrive/ngx-utils",
-	"version": "18.0.1",
+	"version": "18.1.0",
   "description": "A series of abstracts, utils, pipes and services for Angular applications.",
   "keywords": [
```
IbenTesara, line 5: This should technically be done in a separate PR (sorry 😬 ) but could you update the keywords here with references to the service you added?

IbenTesara (approved):

PR #184, opened 2024-07-19
feat(ngx-utils): add a class to the withRouterLinks substitute element

WHeirstrate:
This addition will allow for a custom class to be defined onto the host of the new element through the injection context and the pipe arguments.


IbenTesara (approved):
I'm not entirely convinced by the term "hostClass" because it might cause some confusion, but honestly I can't for the life of me think of a better name for it, so hostClass it is!

PR #186, opened 2024-07-19
build(ngx-utils): v18.2.0

WHeirstrate:


libs/utils/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@studiohyperdrive/ngx-utils",
-	"version": "18.1.0",
+	"version": "18.1.1",
```
IbenTesara, line 3: Now just change it here as well and you're done :D

WHeirstrate: DONE 🤦

PR #192, opened 2024-08-21
feat(ngx-tour): add and remove a body class to allow for custom stying when the tour is active

WHeirstrate:
# Description
Upon starting the tour, a `ngx-tour-active` class is applied to the body. This allows for styling customisation like disabling pointer-events, for example.

# Screenrecording
https://github.com/user-attachments/assets/c7bd951f-9a54-40d7-811d-4161a3b9a0a4

### DIFF
--- a/apps/layout-test/src/pages/main/main.component.html
+++ b/apps/layout-test/src/pages/main/main.component.html
@@ -1,12 +1,10 @@
-<h1>
-	NgxTour
-</h1>
+<h1>NgxTour</h1>
 
 <button (click)="startTour()">
-	{{dataString$ | async}}
+	{{ dataString$ | async }}
 </button>
 
-<h1 tourItem="start">NgxDisplayContent</h1>
+<h1 tourItem="start" contenteditable="true">NgxDisplayContent</h1>
 <div>
 	<p *displayContent="form.value; configuration: { error: { data: 'I am an error' } }">
 		Hello this is the content we wish to show!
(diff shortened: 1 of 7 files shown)


apps/layout-test/src/pages/main/main.component.html
```diff
@@ -1,12 +1,10 @@
...
-</h1>
+<h1>NgxTour</h1>
 
 <button (click)="startTour()">
-	{{dataString$ | async}}
+	{{ dataString$ | async }}
 </button>
 
-<h1 tourItem="start">NgxDisplayContent</h1>
+<h1 tourItem="start" contenteditable="true">NgxDisplayContent</h1>
```
IbenTesara, line 7: Why was this added?

IbenTesara:
Can you also fix the tests?

apps/layout-test/src/pages/main/main.component.html
WHeirstrate: I added this to show that the body class with `pointer-events: none` is properly set onto the body only when the tour is active, as can be seen in the screenrecording.
I found this the easiest way to showcase this without adding another tour step. So just laziness, I guess? 🤷

DenisValcke (approved):

WHeirstrate:
The tests will be fixed in a later PR. Currently, we are unable to find a solution as to why they are failing.
cc: @IbenTesara

IbenTesara (approved):

PR #195, opened 2024-08-22
feat(ngx-utils): Add replace-elements pipe

IbenTesara:
A more generic version of the WithRouterLinks pipe that now allows for any kind of WebComponent to be added

### DIFF
--- a/libs/utils/src/lib/pipes/replace-elements/replace-elements.pipe.ts
+++ b/libs/utils/src/lib/pipes/replace-elements/replace-elements.pipe.ts
@@ -1,0 +1,83 @@
+import { Inject, Pipe, PipeTransform } from '@angular/core';
+import { DomSanitizer, SafeHtml } from '@angular/platform-browser';
+
+import { NgxReplaceElementsConfigurationToken } from '../../tokens';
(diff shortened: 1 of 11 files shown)


libs/utils/src/lib/pipes/replace-elements/replace-elements.pipe.ts
```diff
@@ -0,0 +1,83 @@
...
+
+			// Iben: Create a new element within the Document based on the provided selector.
+			// The selector can be any native or custom web component (not an Angular component).
+			// Keep in mind that the element will need to have a lowercase input prop for the reference.
+			const replacement: HTMLElement = body.createElement(element);
+
+			// Iben: If the item included data, we set these attributes
+			if (item.data) {
+				Object.entries(item.data).forEach(([key, value]) => {
+					replacement.setAttribute(key, value);
```
DenisValcke, line 67: Keep in mind that HTML attributes are always lower cased. Maybe this should be documented somewhere to avoid people using camelCases properties here and not understanding why it isn't working.

https://developer.mozilla.org/en-US/docs/Web/API/Element/setAttribute
> The attribute name is automatically converted to all lower-case when setAttribute() is called on an HTML element in an HTML document.

DenisValcke (approved):

PR #207, opened 2024-08-29
feat(ngx-inform): Setup ngx-inform package

IbenTesara:
Setup for the `ngx-inform` package, which will be the home for tooltips, toasts and snackbars.

### DIFF
--- a/libs/inform/package.json
+++ b/libs/inform/package.json
@@ -1,0 +1,31 @@
+{
+	"name": "@studiohyperdrive/ngx-inform",
+	"version": "18.0.0",
+  "description": "A lightweight ARIA compliant customizable approach for common and complex inform flows in Angular.",
+  "keywords": [
+    "angular",
+    "angular2",
+    "inform",
+    "ARIA",
+    "tooltip",
+    "tip",
+    "accessibility"
+  ],
+  "homepage": "https://github.com/studiohyperdrive/ngx-tools/tree/master/libs/inform",
+  "license": "MIT",
+  "author": {
+    "name": "Studio Hyperdrive",
+    "url": "https://studiohyperdrive.be/"
+  },
+  "repository": {
+    "type": "git",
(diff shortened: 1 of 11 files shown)


libs/inform/package.json
```diff
@@ -0,0 +1,30 @@
...
+	"name": "@studiohyperdrive/ngx-inform",
+	"version": "18.0.0",
+  "description": "A lightweight ARIA compliant customizable approach for common and complex inform flows in Angular.",
+  "keywords": [
+    "angular",
+    "angular2",
+    "inform",
+    "ARIA",
+    "tooltip",
+    "tip"
```
WHeirstrate, line 11: Accessibility

tsconfig.base.json
```diff
@@ -22,13 +22,14 @@
 			"@ngx/layout": ["libs/layout/src/public-api.ts"],
 			"@ngx/store": ["libs/store/src/public-api.ts"],
 			"@ngx/table": ["libs/table/src/public-api.ts"],
+			"@ngx/tour": ["libs/tour/src/index.ts"],
 			"@ngx/utils": ["libs/utils/src/public-api.ts"],
-			"@ngx/tour": ["libs/tour/src/index.ts"]
```
DenisValcke, line 26: Why does this one not match the others?

WHeirstrate (approved):

tsconfig.base.json
IbenTesara: That's the new approach of Nx, it has been like that for quite a while now (was already the case with the ngx-tour package). I don't know if there's a migration script for it, but I'm hesitant to edit this if this is their new approach 😅

DenisValcke: Fair enough, maybe we should create a task on the board to do the same for the existing packages.

DenisValcke (approved):

PR #209, opened 2024-08-30
feat(ngx-inform): Add ability to disable the tooltip directive

IbenTesara:


libs/inform/src/lib/directives/tooltip/tooltip.directive.ts
```diff
@@ -58,9 +59,9 @@ export class NgxTooltipDirective {
 	@HostBinding('tabIndex') private readonly index = 0;
 
 	/**
-	 * The id of the tooltip, required for accessibility
+	 * The id of the tooltip, required for accessibility. By default, this is an autogenerated UUID.
```
WHeirstrate, line 62: I'd add, like in the documentation, `This id most always be unique`. Then we've done our job to create enough awareness.

WHeirstrate (approved):

PR #216, opened 2024-09-11
feat(ngx-i18n): Make the package cohesive with other packages and add SetLanguageGuard

IbenTesara:
This PR does two things.

1. It renames the services, guards, etc. to be more cohesive with the other packages. Now it is clear they're part of the ngx-i18n package. I at first wanted to write a migration script, but unfortunately did not have the time.
2. It adds a new guard that will automatically set the language 
3. It makes the root I18nService clearer

### DIFF
--- a/libs/i18n/src/lib/services/root-i18n/root-i18n.service.ts
+++ b/libs/i18n/src/lib/services/root-i18n/root-i18n.service.ts
@@ -1,20 +1,63 @@
-import { Injectable } from '@angular/core';
+import { isPlatformBrowser } from '@angular/common';
+import { Inject, Injectable, PLATFORM_ID } from '@angular/core';
 import { BehaviorSubject, Observable } from 'rxjs';
 
+import { NgxI18nConfiguration } from '../../i18n.types';
+import { NgxI18nConfigurationToken } from '../../tokens';
+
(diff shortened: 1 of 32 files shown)


libs/i18n/src/lib/services/root-i18n/root-i18n.service.ts
```diff
@@ -1,20 +1,63 @@
...
 
-	public setCurrentLanguage(language: string) {
+	/**
+	 * Sets the current language of the application and saves it to the local storage. Returns true if the language was set
+	 *
+	 * @param language - The provided language
+	 */
+	public setCurrentLanguage(language: string): void {
+		// Iben: If a language is set that's not part of the available languages, we return a warn
+		if (!this.configuration.availableLanguages.includes(language)) {
```
DenisValcke, line 44: Is it safe to imply that `configuration` will be defined and have `availableLanguages` as property + is it safe to assume that that property will always be Array?

DenisValcke (approved):

libs/i18n/src/lib/services/root-i18n/root-i18n.service.ts
IbenTesara: As mentioned in private:

- If the configuration isn't defined it isn't provided and it will return a Injection error
- The configuration is typed and cannot not contain availableLanguages, so unless someone uses any this should also not be an issue

PR #217, opened 2024-09-12
fix(ngx-i18n): Add initializeLanguage

IbenTesara:
The SetLanguageGuard takes care of initializing the language when no language was set and this works perfectly when the language is not part of the route yet.

However, when the language parameter was already set, the guard is not called and the I18nGuard did not initialize the language. This PR fixes the issue.

### DIFF
--- a/libs/i18n/src/lib/services/root-i18n/root-i18n.service.ts
+++ b/libs/i18n/src/lib/services/root-i18n/root-i18n.service.ts
@@ -40,24 +40,68 @@ export class NgxI18nRootService {
 	 * @param language - The provided language
 	 */
 	public setCurrentLanguage(language: string): void {
-		// Iben: If a language is set that's not part of the available languages, we return a warn
+		// Iben: get the new language
+		const newLanguage = this.getNewLanguage(language);
+
+		// Iben: Save the current language to the localStorage when we're in the browser
+		if (isPlatformBrowser(this.platformId)) {
+			localStorage.setItem('ngx-i18n-language', newLanguage);
+		}
+
+		// Iben: Update the subject
+		this.currentLanguageSubject.next(newLanguage);
+	}
+
+	/**
+	 * Sets the initial language of the application when no language is set yet.
+	 *
+	 * If a previous language was set in the local storage, said language is used. If not, the default language gets used.
+	 */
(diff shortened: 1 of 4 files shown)


libs/i18n/src/lib/services/root-i18n/root-i18n.service.ts
```diff
@@ -40,24 +40,56 @@ export class NgxI18nRootService {
...
+			if (this.currentLanguage) {
+				return;
+			}
+
+			// Iben: If no language exists, we use the default language
+			newLanguage = this.configuration.defaultLanguage;
 		}
 
 		// Iben: Save the current language to the localStorage when we're in the browser
 		if (isPlatformBrowser(this.platformId)) {
```
DenisValcke, line 65: Sidenote: We should look into how to share functionalitity between packages because this would be better handled with the WindowService from the utils.

libs/i18n/src/lib/services/root-i18n/root-i18n.service.ts
```diff
@@ -40,24 +40,56 @@ export class NgxI18nRootService {
...
-			// Iben: Early exit
-			return;
+			// Iben: If there is already a language set, we early exit and keep the remaining language
+			if (this.currentLanguage) {
+				return;
+			}
+
+			// Iben: If no language exists, we use the default language
+			newLanguage = this.configuration.defaultLanguage;
 		}
```
DenisValcke, line 62: I would move the functionality of this `if (!this.configuration.availableLanguages.includes(language)) { ... }` to a seperate private method `getNewLanguage` or something to isolate this functionality from setting it.

libs/i18n/src/lib/services/root-i18n/root-i18n.service.ts
IbenTesara: Yes, this is something mentioned in the comment at the top of the service, but we should really start looking into a solution to this.

libs/i18n/src/lib/services/root-i18n/root-i18n.service.ts
IbenTesara: Good call!

PR #219, opened 2024-09-18
feat(ngx-inform): Add modal setup

IbenTesara:
Adds a forced WCAG/ARIA compliant wrapper around the CDK Dialog.


libs/inform/src/lib/abstracts/modal/modal.abstract.component.ts
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * An abstract for the NgxModalService
+ */
+@Directive()
+export class NgxModalAbstractComponent<ActionType extends string = string, DataType = any> {
+	/**
+	 * Remove the modal on escape pressed
+	 */
+	@HostListener('document:keydown.escape') onEscape() {
```
DenisValcke, line 11: This can be a private member.

libs/inform/src/lib/services/modal/modal.service.ts
```diff
@@ -0,0 +1,234 @@
...
+			return NEVER;
+		}
+
+		// Iben: Render the modal
+		const modal = this.createModalComponent<ActionType>(options, component);
+
+		// Iben: Return the modal action
+		return combineLatest([
+			// Iben: Set the start value to undefined so both actions at least emit once
+			modal.action.pipe(startWith(undefined)),
```
DenisValcke, line 68: I think we've discussed this in the past, but just noting that using `undefined` is only considerd when using ES5+ in strict mode. If not, `undefined` can be overwritten and this can cause unwanted behaviour.

(Also applies to several other places in this PR)

libs/inform/src/lib/services/modal/modal.service.ts
```diff
@@ -0,0 +1,234 @@
...
+			modal.close.pipe(
+				// Iben: Map so we can keep the emit value void, but can work with the filter later down the line
+				map(() => 'NgxModalClose'),
+				// Iben: Set the start value to undefined so both actions at least emit once
+				startWith(undefined)
+			),
+		]).pipe(
+			// Iben: Only emit if one of the two actions actually has an emit
+			filter(([action, closed]) => {
```
DenisValcke, line 77: I would add interfaces in these arrow functions (filter, map, tap, ...) for clarity.

libs/inform/src/lib/services/modal/modal.service.ts
```diff
@@ -0,0 +1,234 @@
...
+			takeUntil(this.modalClosedSubject)
+		);
+	}
+
+	/**
+	 * Closes the currently active modal
+	 *
+	 * * @param onClose - An optional onClose function
+	 */
+	public close(onClose?: Function): void {
```
DenisValcke, line 110: I would specify this callback function. No arguments and no return value so change it to `() => void`.

libs/inform/src/lib/services/tooltip/tooltip.service.ts
```diff
@@ -170,11 +170,13 @@ export class NgxTooltipService implements OnDestroy {
 	 * Removes the tooltip.
 	 */
 	public removeToolTip() {
```
DenisValcke, line 172: Provide a return type.

libs/inform/src/lib/services/modal/modal.service.ts
```diff
@@ -0,0 +1,234 @@
...
+} from 'rxjs';
+
+import { NgxModalConfiguration, NgxModalOptions } from '../../types';
+import { NgxModalConfigurationToken } from '../../tokens';
+import { NgxModalAbstractComponent } from '../../abstracts';
+
+/**
+ * A wrapper service to Angular CDK Dialog that provides a WCAG/ARIA compliant implementation of modals
+ *
+ * @export
```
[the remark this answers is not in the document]
IbenTesara: It's been a thing we've started doing in the VLAIO project a while back and it is in line with what Angular does for their services as well. 

Other than extra information I don't think there's a major benefit, but I don't see a downside either.

libs/inform/src/lib/services/modal/modal.service.ts
IbenTesara: I am aware, but given that this is just internal use I don't see an issue here in this specific spot? Unless you can point me to a potential issue here?

For me it's the same as using `first` vs `take(1)` in rjxs. As long as you know the difference and you use the correct one in the correct situation, I'm fine with either one if you use it in a situation where they're interchangeable.

WHeirstrate (approved):

PR #227, opened 2024-09-20
fix(modalService): replace the subject with a behaviorSubject and document entities

WHeirstrate:
The subscription got closed before emitting. This is resolved, along with added documentation of the modal interfaces.

https://github.com/user-attachments/assets/8d056ec1-6fb2-48d0-8046-5a78047917cc

### DIFF
--- a/libs/inform/src/lib/services/modal/modal.service.ts
+++ b/libs/inform/src/lib/services/modal/modal.service.ts
@@ -1,13 +1,13 @@
 import { Inject, Injectable, Optional, Type } from '@angular/core';
 import { Dialog } from '@angular/cdk/dialog';
 import {
+	BehaviorSubject,
 	combineLatest,
 	filter,
 	map,
 	NEVER,
 	Observable,
 	startWith,
-	Subject,
 	takeUntil,
 	tap,
 } from 'rxjs';
@@ -25,9 +25,14 @@ import { NgxModalAbstractComponent } from '../../abstracts';
 @Injectable({ providedIn: 'root' })
 export class NgxModalService {
 	/**
-	 * A subject to hold the close event
+	 * A subject that keeps track of whether a modal is currently active
 	 */
-	private modalClosedSubject: Subject<void>;
+	private hasModalSubject: BehaviorSubject<boolean> = new BehaviorSubject(false);
+
+	/**
+	 * An observable that keeps track of whether a modal is currently active.
+	 */
+	public readonly hasActiveModal$: Observable<boolean> = this.hasModalSubject.asObservable();
 
 	constructor(
 		@Optional()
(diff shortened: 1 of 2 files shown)


libs/inform/src/lib/services/modal/modal.service.ts
```diff
@@ -122,9 +126,9 @@ export class NgxModalService {
 		this.dialogService.closeAll();
 
 		// Iben: Mark the modal as closed
-		this.modalClosedSubject.next();
-		this.modalClosedSubject.complete();
-
+		setTimeout(() => {
```
IbenTesara, line 129: Better add a comment why we use the setTimeOut here

IbenTesara (changes requested):

libs/inform/src/lib/services/modal/modal.service.ts
```diff
@@ -25,9 +25,14 @@ import { NgxModalAbstractComponent } from '../../abstracts';
 @Injectable({ providedIn: 'root' })
 export class NgxModalService {
 	/**
-	 * A subject to hold the close event
+	 * A subject that keeps track of whether a modal is currently active
 	 */
-	private modalClosedSubject: Subject<void>;
+	private hasModalSubject: BehaviorSubject<boolean> = new BehaviorSubject(false);
```
DenisValcke, line 30: Don't forget the `$` suffix to indicate an rxjs stream.

libs/inform/src/lib/services/modal/modal.service.ts
```diff
@@ -44,25 +49,23 @@ export class NgxModalService {
 	public open<ActionType extends string = string, DataType = any>(
 		options: NgxModalOptions<ActionType, DataType>
 	): Observable<ActionType> {
-		// Iben: If there still is an active subject running, the modal is still active and we early exit.
-		if (this.modalClosedSubject !== undefined && !this.modalClosedSubject?.closed) {
+		// Iben: If a previous modal is still active, we early exit.
+		if (this.hasModalSubject.value) {
```
DenisValcke, line 53: Not a big fan of the `.value` syntax because it gets the current value synchronously from an inherently asynchronous stream. Not telling you to not do this if this is the best way to go about it, but keep in mind that this can cause unwanted side effects. Only use the `.value` property when there is no other way to go about it.

DenisValcke: Sidenote: wouldn't it be better if the options had a `hasPriority` value or something that can indicate if the new modal should close the previous one? I can imagine that your second modal getting "lost" because of an other one still being open, might be unwanted behaviour.

IbenTesara: Whilst I understand your concerns with the value call here, in this case you just want to know the current value and not respond every time the value updates. A subscription here would mean you'd place a `take(1)`, and at that point it becomes the same as requesting the current value with the value call. I would even go as far as to say that the value approach is better in that case, because you don't start a subscription when it's not needed.

When it comes to the second remark, is that something other modal packages also do? 🤔 AFAIK all packages I've ever used just don't allow you to add a second modal if there's already one open.

libs/inform/src/lib/services/modal/modal.service.ts
DenisValcke: I always use it for any type of rxjs stream. I don't know what the convention is for the codebase, I don't have a strong preference, but let's avoid having Frankenstein code with two or more conventions for the same thing.

libs/inform/src/lib/services/modal/modal.service.ts
DenisValcke: I think it would just be a nice to have to allow for certain modals to have a higher priority and thus closing a previous one. Maybe a nice to have feature for the board, but something to think about.

DenisValcke (approved):

IbenTesara (approved):

PR #231, opened 2024-09-25
feat(ngx-table): Make the table accessible through the tree-grid pattern

IbenTesara:
Adds WCAG compliance through the [treegrid pattern. ](https://www.w3.org/WAI/ARIA/apg/patterns/treegrid/)

Some form of Cypress/other related testing should be added in the future to properly test the keyboard interactions.

### DIFF
--- a/libs/table/src/lib/directives/tree-grid/tree-grid-row.directive.ts
+++ b/libs/table/src/lib/directives/tree-grid/tree-grid-row.directive.ts
@@ -1,0 +1,272 @@
+import {
+	Directive,
+	ElementRef,
+	EventEmitter,
+	HostBinding,
+	HostListener,
+	Input,
+	OnDestroy,
+	Output,
+} from '@angular/core';
+
+import { NgxTreeGridCellTarget, NgxTreeGridRowTarget } from '../../interfaces';
+import { NgxHasFocusDirective } from '../has-focus-action';
+import { NgxTreeGridDirective } from './tree-grid.directive';
+import { NgxTreeGridCellDirective } from './tree-grid.cell.directive';
+
+/**
+ * A row directive to handle navigation according to the WCAG treegrid pattern
+ *
+ * See https://www.w3.org/WAI/ARIA/apg/patterns/treegrid/
+ */
+@Directive({
+	selector: '[ngxTreeGridRow]',
+	standalone: true,
+})
+export class NgxTreeGridRowDirective extends NgxHasFocusDirective implements OnDestroy {
+	/**
+	 * An array of all cells in the row
+	 */
+	private readonly cells: NgxTreeGridCellDirective[] = [];
+
+	/**
+	 * Sets focus on to the previous row
+	 */
+	@HostListener('keydown.ArrowUp') private moveUp(): void {
+		this.handleWhenFocussed(() => {
+			this.parent.getRow(this.ngxTreeGridRow - 1)?.focus();
+		});
+	}
+
+	/**
+	 * Sets focus to the next row
+	 */
+	@HostListener('keydown.ArrowDown') private moveDown(): void {
+		this.handleWhenFocussed(() => {
+			this.parent.getRow(this.ngxTreeGridRow + 1)?.focus();
+		});
+	}
+
+	/**
+	 * Sets the focus on the first cell in the row or closes the row when expanded
+	 */
+	@HostListener('keydown.ArrowRight') private arrowRight(): void {
+		this.handleWhenFocussed(() => {
+			// Iben: Only when the row is not expanded we can move to the next cell, as by WCAG design
+			if (this.parent.ngxTreeGridExpandable && this.ngxTreeGridRowExpanded) {
+				this.ngxTreeGridRowExpandRow.emit(false);
+				return;
+			}
+
+			// Iben: Focus on the first cell
+			this.cells[0]?.focus();
+		});
+	}
+
+	/**
+	 * Expand the row when closed
+	 */
+	@HostListener('keydown.ArrowLeft') private arrowLeft(): void {
+		this.handleWhenFocussed(() => {
+			this.ngxTreeGridRowExpandRow.emit(true);
+		});
+	}
+
+	/**
+	 * Sets focus on the first row of the grid
+	 */
+	@HostListener('keydown.PageUp') private moveToTopPageUp(): void {
+		this.moveToTop();
+	}
+
+	/**
+	 * Sets focus on the first row of the grid
+	 */
+	@HostListener('keydown.Home') private moveToTopHome(): void {
+		this.moveToTop();
+	}
+
+	/**
+	 * Sets focus on the first row of the grid
+	 */
+	@HostListener('keydown.control.Home') private moveToTopControlHome(): void {
+		this.moveToTop();
+	}
+
+	/**
+	 * Sets focus on the last row of the grid
+	 */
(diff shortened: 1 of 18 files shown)


libs/table/src/lib/table/ngx-table.component.ts
```diff
@@ -441,16 +461,21 @@ export class NgxTableComponent
 			}
 		}
 
+		// Iben: Handle the row state
+		this.handleRowState(index, !this.openRows.has(index) ? 'open' : 'close');
+	}
+
+	public handleRowState(index: number, action: 'open' | 'close') {
```
WHeirstrate, line 468: Documentation and return type

WHeirstrate (changes requested):

WHeirstrate (approved):

libs/table/src/lib/directives/tree-grid/tree-grid-row.directive.ts
```diff
@@ -0,0 +1,272 @@
...
+			this.parent.getRow(this.ngxTreeGridRow + 1)?.focus();
+		});
+	}
+
+	/**
+	 * Emits a select event to select a row above the current row and move focus to that row
+	 */
+	@HostListener('keydown.shift.ArrowUp') selectPreviousRow(): void {
+		this.handleWhenFocussed(() => {
+			this.parent.getRow(this.ngxTreeGridRow + 1)?.selectRow();
```
DenisValcke, line 150: Is this `+` correct?

libs/table/src/lib/pipes/aria-sort/aria-sort.pipe.ts
```diff
@@ -0,0 +1,23 @@
...
+		currentSorting: NgxTableSortEvent;
+		cell: NgxAbstractTableCellDirective;
+	}): 'none' | 'ascending' | 'descending' {
+		const { cell } = value;
+
+		if (!cell || !cell.sortDirection) {
+			return 'none';
+		}
+
+		return cell.sortDirection.toLocaleLowerCase() as 'ascending' | 'descending';
```
DenisValcke, line 21: Maybe put this union type in the tree-grid types, for convenience?

libs/table/src/lib/interfaces/treegrid.types.ts
DenisValcke, line 1: This is the only file that is written `treegrid` instead of `tree-grid`, small typo.

libs/table/src/lib/directives/tree-grid/tree-grid.cell.directive.ts
```diff
@@ -0,0 +1,179 @@
...
+export class NgxTreeGridCellDirective extends NgxHasFocusDirective {
+	/**
+	 * The parent row of the cell
+	 */
+	private row: NgxTreeGridRowDirective;
+
+	/**
+	 * Marks the cell as focusable, but only by setting its focus programmatically, not by the tab key
+	 */
+	@HostBinding('attr.tabIndex') private canFocus: number = -1;
```
DenisValcke, line 26: This can also be moved to the `host` property of the decorator. A matter of preference, but since it's a static value, might be something to consider.

libs/table/src/lib/directives/tree-grid/tree-grid.cell.directive.ts
```diff
@@ -0,0 +1,179 @@
...
+
+	/**
+	 * Set focus on the previous cell to the left
+	 */
+	@HostListener('keydown.ArrowLeft', ['$event']) public moveLeft(event: Event): void {
```
DenisValcke, line 31: Just out of interest, why is the `ArrowLeft` listener so much more complicated than the others?

(Might be good to also clarify in the comment)

libs/table/src/lib/directives/tree-grid/tree-grid.cell.directive.ts
```diff
@@ -0,0 +1,179 @@
...
+	 * Searches for a focusable element in the cell
+	 */
+	private findFocusableElement(): HTMLElement | undefined {
+		let result: HTMLElement;
+
+		// Iben: Loop over each first-level element of the children
+		for (let element of [...this.elementRef.nativeElement.children]) {
+			if (!result) {
+				// Iben: Check if we can focus on the element
+				element.focus();
```
DenisValcke, line 159: Does something need to happen if we can't? Just asking because you speak of "check if we can".

libs/table/src/lib/directives/tree-grid/tree-grid.directive.ts
```diff
@@ -0,0 +1,102 @@
...
+	selector: '[ngxTreeGrid]',
+	standalone: true,
+})
+export class NgxTreeGridDirective {
+	/**
+	 * An array of all rows in the treegrid
+	 */
+	private rows: NgxTreeGridRowDirective[] = [];
+
+	@HostBinding('attr.role') private role: 'table' | 'treegrid' = 'table';
```
DenisValcke, line 21: Maybe put this union type in the tree-grid types, for convenience?

libs/table/src/lib/directives/tree-grid/tree-grid.directive.ts
```diff
@@ -0,0 +1,102 @@
...
+	 *
+	 * @param cell - The provided cell
+	 */
+	public registerCell(index: number, cell: NgxTreeGridCellDirective) {
+		this.getRow(index)?.registerCell(cell);
+	}
+
+	ngOnChanges() {
+		// Iben: Set the role based on the tree grid
+		this.role = this.ngxTreeGrid ? 'treegrid' : 'table';
```
DenisValcke, line 100: Maybe put this union type in the tree-grid types, for convenience?

libs/table/src/lib/directives/tree-grid/tree-grid-row.directive.ts
IbenTesara: Good catch 👀

libs/table/src/lib/directives/tree-grid/tree-grid.cell.directive.ts
IbenTesara: I've added a reference to it in the comment on the if statement.

Basically, arrowLeft on a row opens the row. But because of how focus works and how the Hostlistener works in Angular; calling arrowLeft here opened the detail row which would mean that the detail row opens whenever you navigate back on the cells; which you don't want.

This is not the case for the other arrow keys.

libs/table/src/lib/pipes/aria-sort/aria-sort.pipe.ts
IbenTesara: The reason why I didn't is because it's not really a type worth sharing everywhere. I prefer to keep these string types in the component/util/etc. where it's used if it has no other value throughout the application/lib.

libs/table/src/lib/directives/tree-grid/tree-grid.directive.ts
IbenTesara: Same as above, there's no use for this type anywhere else, so I prefer to keep it here.

libs/table/src/lib/directives/tree-grid/tree-grid.cell.directive.ts
IbenTesara: If we can't then we focus on the main cell in the other method!

DenisValcke (approved):

PR #233, opened 2024-09-25
fix(ngx-inform): allow modal data set in the modal component to be emitted to the parent along with the type

WHeirstrate:
https://github.com/user-attachments/assets/eec21667-a9f5-42cf-b595-6bf86d0207be

### DIFF
--- a/libs/inform/src/lib/abstracts/modal/modal.abstract.component.ts
+++ b/libs/inform/src/lib/abstracts/modal/modal.abstract.component.ts
@@ -1,10 +1,11 @@
 import { Directive, EventEmitter, HostListener, Input, Output } from '@angular/core';
+import { NgxModalActionType } from '../../types';
 
 /**
  * An abstract for the NgxModalService
  */
 @Directive()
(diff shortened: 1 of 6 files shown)


libs/inform/src/lib/abstracts/modal/modal.abstract.component.ts
```diff
@@ -4,7 +4,10 @@ import { Directive, EventEmitter, HostListener, Input, Output } from '@angular/c
  * An abstract for the NgxModalService
  */
 @Directive()
-export class NgxModalAbstractComponent<ActionType extends string = string, DataType = any> {
+export class NgxModalAbstractComponent<
+	ActionType extends string | { type: string; data: any },
```
IbenTesara, line 8: I would turn `string | {type: string; data: any}` into a type you can just share. 

Something like this:
```
export type NgxModalActionType<StringType extends string = string, DataType = any> = StringType | {type: StringType; data: DataType}
```

IbenTesara (changes requested):

IbenTesara (approved):

PR #248, opened 2024-10-08
Feat/ngx table/add tabindex to table row only when it has details

### LINKED ISSUE #247: [ngx-table]: Table rows are focusable with TAB but not clickable (WCAG)
After a WCAG audit did we get a report for the table component. They suggest removing `tabindex=‘0’` but I wanted to ask you guys first before I open a PR. Was that put on by a reason? Maybe it should be optional.

> Testcriterium: 9.2.4.3 Focus volgorde
> [EN 301 549](https://www.etsi.org/deliver/etsi_en/301500_301599/301549/03.02.01_60/en_301549v030201p.pdf#%5B%7B%22num%22%3A124%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C54%2C417%2C0%5D)
> [WCAG Quick Reference](https://www.w3.org/WAI/WCAG21/quickref/#focus-order)
> [WCAG Understanding](https://www.w3.org/WAI/WCAG21/Understanding/focus-order)
> Type probleem: Breach
> 
> The rows in the table are accessible with TAB, but they are not clickable. Nothing happens when you press ENTER either. This is confusing. The rows in the table are not interactive elements, so they do not need to be accessible with TAB. Remove the tabindex="0" attribute on the tr-lements in the table.

ekincia:
`tabindex` attribute will only be added for rows with `detailRowTemplate` and if it's `selectable` 
resolves #247

### DIFF
--- a/libs/table/src/lib/table/ngx-table.component.html
+++ b/libs/table/src/lib/table/ngx-table.component.html
@@ -189,7 +189,7 @@
 	<tr
 		*cdkRowDef="let row; columns: definedColumns(); let rowIndex = dataIndex"
 		cdk-row
-		tabindex="0"
+		[attr.tabindex]="detailRowTemplate !== undefined || selectable ? 0 : -1"
 		aria-level="1"
 		[attr.aria-setsize]="detailRowTemplate ? 1 : 0"
 		[ngClass]="[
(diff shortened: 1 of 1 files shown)


libs/table/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@studiohyperdrive/ngx-table",
-	"version": "18.1.2",
```
IbenTesara, line 3: This should be part of a separate PR.

IbenTesara (approved):

WHeirstrate (approved):

PR #251, opened 2024-10-16
feat(ngx-layout): Add accordion component

IbenTesara:
afbeelding

### DIFF
--- a/libs/layout/src/lib/components/accordion/item/accordion-item.component.scss
+++ b/libs/layout/src/lib/components/accordion/item/accordion-item.component.scss
@@ -1,0 +1,17 @@
+:host {
+	details > summary {
+		list-style: none;
+	}
+
+	details > summary::-webkit-details-marker {
+		display: none;
+	}
+
+	summary {
+		cursor: pointer;
+
+		&.is-disabled {
+			cursor: not-allowed;
+		}
+	}
+}
--- a/libs/layout/src/lib/components/accordion/item/accordion-item.component.ts
+++ b/libs/layout/src/lib/components/accordion/item/accordion-item.component.ts
@@ -1,0 +1,201 @@
+import {
+	AfterViewInit,
+	ChangeDetectionStrategy,
+	ChangeDetectorRef,
+	Component,
+	ContentChild,
+	ElementRef,
+	HostListener,
+	Input,
+	OnDestroy,
+	OnInit,
+	Renderer2,
+	TemplateRef,
+	ViewChild,
+} from '@angular/core';
+import { UUID } from 'angular2-uuid';
+
+import { NgTemplateOutlet } from '@angular/common';
+import { NgxAccordionComponent } from '../accordion.component';
+
+/**
+ * A WCAG/ARIA compliant implementation of an item in the accordion pattern.
+ *
+ * https://www.w3.org/WAI/ARIA/apg/patterns/accordion/
+ */
+@Component({
+	selector: 'ngx-accordion-item',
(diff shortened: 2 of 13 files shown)


libs/layout/src/lib/components/accordion/item/accordion-item.component.scss
```diff
@@ -0,0 +1,17 @@
+:host {
+	details > summary {
+		list-style: none;
+	}
+
+	details > summary::-webkit-details-marker {
```
WHeirstrate, line 6: Should there not be any forced indication that this element is clickable / expandable? By removing this marker, this could very well be just a couple of plain text sentences. Personally, his does not feel very accessible.

libs/layout/src/lib/components/accordion/item/accordion-item.component.ts
```diff
@@ -0,0 +1,201 @@
...
+		// Iben: Prevent the accordion from being opened if it is disabled
+		this.renderer.listen(this.detailsElement.nativeElement, 'click', (event: Event) => {
+			if (this.disabled) {
+				event.preventDefault();
+				event.stopImmediatePropagation();
+			}
+		});
+
+		// Iben: Listen to the open state of details and update the internal one
+		this.renderer.listen(this.detailsElement.nativeElement, 'toggle', (event: ToggleEvent) => {
```
WHeirstrate, line 175: Should these listeners not be removed in the `ngOnDestroy`? And if it isn't required due to the destruction of the `Renderer2` class, shouldn't we explicitly remove the listeners in the same way we usually also push `void` onto the `destroyed$` observable before completion?

libs/layout/src/lib/components/accordion/item/accordion-item.component.scss
IbenTesara: It's not mandatory according to the WCAG guidlines no. [And if you do, those should also be aria-hidden. ](https://www.a11y-collective.com/blog/accessible-accordion/) So in terms of accessibility there's no issue here, the details element already provides enough accessibility information for end-users.

The templates allow you to add the marker yourself, so we have to remove it here. Especially given that some designs will want it at the end or use a completely different icon.

libs/layout/src/lib/components/accordion/item/accordion-item.component.ts
IbenTesara: There is no removeListener method nor does the listen method return a listener to cancel. All I've found in terms of recent information is that once your component gets destroyed (and the instance of Renderer2 along with it), all the listeners get removed as well. In line with how we don't need to remove HostListeners manually either.

It's a good question, I asked myself the same one when implementing it 😂

WHeirstrate: Makes sense! 👍

libs/layout/src/lib/components/accordion/item/accordion-item.component.scss
WHeirstrate: Yeah, that's fair reasoning. Would it be too opinionated to encourage users to provide an indicator in the documentation?

IbenTesara: I don't really think it's _really_ necessary, I think the vast majority of users will just use an indicator either way. If it was strongly suggested in terms of accessibility I would agree, but given that it's not, it's fair game to add it or not.

WHeirstrate (approved):

PR #253, opened 2024-10-17
feat(utils): the utils now contain a service that wraps around the BroadcastChannel API

DenisValcke:
This PR adds the `BroadcastChannelService` to the utils. This service wraps around the BroadcastChannel API:
https://developer.mozilla.org/en-US/docs/Web/API/Broadcast_Channel_API

### DIFF
--- a/libs/utils/src/lib/services/broadcast-channel/broadcast-channel.md
+++ b/libs/utils/src/lib/services/broadcast-channel/broadcast-channel.md
@@ -28,10 +28,10 @@ The `initChannel` uses the `isPlatformBrowser` check to ensure it only runs in t
 
 If the channel already exists, it will return the existing channel to avoid overriding existing channels and listeners.
 
-If a name is not provided, it will early return and log a warning:
+If a name is not provided, it will early return and log an error:
 
 ```
-channelName is required
+NgxUtils: There was an attempt to initialize a BroadcastChannel without providing a name.
 ```
 
 ### closeChannel
@@ -58,13 +58,7 @@ export class YourComponent {
 
 #### Safety
 
-If the channel does not exist on the Record, it will early return.
-
-If a name is not provided, it will early return and log a warning:
-
-```
-channelName is required
-```
+If the channel does not exist on the Record or the name is not provided, it will early return.
 
 ### postMessage
 
@@ -95,23 +89,17 @@ export class YourComponent {
 
 #### Safety
 
-If the channel does not exist on the Record, it will early return and log a warning to give a notice that the channel has not been initialized.
+If the channel does not exist on the Record or the name is not provided, it will early return and log an error to give a notice.
 
 ```
-BroadcastChannel not initialized, message not sent
+NgxUtils: There was an attempt to post a message to a channel without providing a name or the selected channel does not exist. The included message was:
 ```
 
 _This warning will include the message that has been included to give a better understanding of what message was not sent._
 
-If a name is not provided, it will early return and log a warning.
+### selectChannelMessages
 
-```angular2html
-channelName is required
-```
-
-### selectChannel
-
-The `selectChannel` method will return a subscription wrapped around the `message` event of the channel with the given name.
+The `selectChannelMessages` method will return a subscription wrapped around the `message` event of the channel with the given name.
 
 ```typescript
 import { NgxBroadcastChannelService } from '@studiohyperdrive/ngx-utils';
@@ -123,7 +111,7 @@ export class YourComponent {
 		// Open up a channel for this component OnInit.
 		this.broadcastChannelService.initChannel('your-channel-name');
 
-		this.broadcastChannelService.selectChannel('your-channel-name').subscribe({
+		this.broadcastChannelService.selectChannelMessages('your-channel-name').subscribe({
 			// Handle the message event.
 			next: (message: MessageEvent) => {
 				console.log(message.data);
@@ -149,8 +137,54 @@ export class YourComponent {
 
 #### Safety
 
-If a name is not provided, it will early return an `EMPTY` and log a warning.
+If the channel does not exist on the Record or the name is not provided, it will early return an `EMPTY` and log an error.
 
 ```angular2html
-channelName is required
+NgxUtils: There was an attempt to select a BroadcastChannel's messages without providing a name or the selected channel does not exist.
+```
+
+### selectChannelMessageErrors
+
+The `selectChannelMessageErrors` method will return a subscription wrapped around the `message` event of the channel with the given name.
+
+```typescript
+import { NgxBroadcastChannelService } from '@studiohyperdrive/ngx-utils';
+
+export class YourComponent {
+	constructor(private readonly broadcastChannelService: NgxBroadcastChannelService) {}
+
+	public ngOnInit(): void {
+		// Open up a channel for this component OnInit.
+		this.broadcastChannelService.initChannel('your-channel-name');
+
+		this.broadcastChannelService.selectChannelMessageErrors('your-channel-name').subscribe({
+			// Handle the message event.
+			next: (messageError: MessageEvent) => {
+				console.log(messageError.data);
+			},
+			// When the channelName is not provided, an EMPTY is returned to not break the subscription.
+			complete: () => {
+				console.log('No channelName provided to the selectChannel method');
+			},
+		});
+	}
+
+	public ngOnDestroy(): void {
+		// Close the created channel OnDestroy.
+		this.broadcastChannelService.closeChannel('your-channel-name');
+	}
+
+	public sendContextMessage(message: string): void {
+		// Send a message through the channel.
+		this.broadcastChannelService.postMessage('your-channel-name', message);
+	}
+}
+```
+
+#### Safety
+
+If the channel does not exist on the Record or the name is not provided, it will early return an `EMPTY` and log an error.
+
+```angular2html
+NgxUtils: There was an attempt to select a BroadcastChannel's message errors without providing a name or the selected channel does not exist.
 ```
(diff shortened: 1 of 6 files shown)


libs/utils/src/lib/services/broadcast-channel/broadcast-channel.service.ts
```diff
@@ -0,0 +1,95 @@
...
+
+	/**
+	 * initChannel
+	 *
+	 * The initChannel method initializes a new BroadcastChannel instance.
+	 *
+	 * @param channelName{string} - The name of the Broadcast Channel.
+	 */
+	public initChannel(channelName: string): void {
+		if (!channelName) {
```
IbenTesara, line 21: This check is done here everywhere, but besides this single instance, is it really necessary?

libs/utils/src/lib/services/broadcast-channel/broadcast-channel.service.ts
```diff
@@ -0,0 +1,95 @@
...
+			console.warn('channelName is required');
```
IbenTesara, line 22: This should in my opinion return an error and not a warning.

The error texts also aren't really in line with how we do it in other packages, we usually start with the package name `NgxUtils:` and then a more human readable error.

libs/utils/src/lib/services/broadcast-channel/broadcast-channel.service.ts
```diff
@@ -0,0 +1,95 @@
...
+
+			return;
+		}
+
+		if (isPlatformBrowser(this.platformId) && !this.broadcastChannel[channelName]) {
```
IbenTesara, line 27: Can we use the WindowService's `runInBrowser` function here?

libs/utils/src/lib/services/broadcast-channel/broadcast-channel.service.ts
```diff
@@ -0,0 +1,95 @@
...
+	 * @param message{any} - The payload to send through the channel.
+	 */
+	public postMessage(channelName: string, message: any): void {
+		if (!channelName) {
+			console.warn('channelName is required');
+
+			return;
+		}
+
+		if (!this.broadcastChannel[channelName]) {
```
IbenTesara, line 69: Here as well, I'd expect an error not a warn. I'd also refer to the initChannel method here.

libs/utils/src/lib/services/broadcast-channel/broadcast-channel.service.ts
```diff
@@ -0,0 +1,95 @@
...
+	/**
+	 * postMessage
+	 *
+	 * The postMessage method sends a message to a selected BroadcastChannel instance.
+	 *
+	 * @param channelName{string} - The name of the Broadcast Channel.
+	 * @param message{any} - The payload to send through the channel.
+	 */
+	public postMessage(channelName: string, message: any): void {
+		if (!channelName) {
```
IbenTesara, line 63: Because you already do this check in the `initChannel` method, this basically already gets handled by the next if check, as it can never be created to begin with.

libs/utils/src/lib/services/broadcast-channel/broadcast-channel.service.ts
```diff
@@ -0,0 +1,95 @@
...
+
+	/**
+	 * subscribeToChannel
+	 *
+	 * The subscribeToChannel method subscribes to the `message` (bc.onmessage) event of a selected BroadcastChannel instance.
+	 *
+	 * @param channelName{string} - The name of the Broadcast Channel.
+	 * @returns Observable<MessageEvent> - The message event of the channel wrapped in an observable.
+	 */
+	public selectChannel(channelName: string): Observable<MessageEvent> {
```
IbenTesara, line 86: Method name does not seem to match with the documentation.

libs/utils/src/lib/services/broadcast-channel/broadcast-channel.service.ts
```diff
@@ -0,0 +1,95 @@
...
+		if (!channelName) {
+			console.warn('channelName is required');
+
+			return EMPTY;
+		}
+
+		return fromEvent<MessageEvent>(this.broadcastChannel[channelName], 'message');
```
IbenTesara, line 93: Shouldn't there be a check if there's a channel?

libs/utils/src/lib/services/broadcast-channel/broadcast-channel.md
```diff
@@ -0,0 +1,156 @@
+# BroadcastChannelService
```
IbenTesara, line 1: Can this also be added to the overall readme?

IbenTesara (changes requested):
A few comments, mostly to keep it in line with the other packages. 

Maybe also update the tags in the package.json?

libs/utils/src/lib/services/broadcast-channel/broadcast-channel.service.ts
```diff
@@ -0,0 +1,95 @@
...
+	}
```
ian-emsens-shd, line 94: Add a similar method for [messageerror](https://developer.mozilla.org/en-US/docs/Web/API/BroadcastChannel/messageerror_event)

libs/utils/src/lib/services/broadcast-channel/broadcast-channel.service.ts
```diff
@@ -0,0 +1,95 @@
...
+	constructor(@Inject(PLATFORM_ID) private platformId: string) {}
+
+	/**
+	 * initChannel
+	 *
+	 * The initChannel method initializes a new BroadcastChannel instance.
+	 *
+	 * @param channelName{string} - The name of the Broadcast Channel.
+	 */
+	public initChannel(channelName: string): void {
```
ian-emsens-shd, line 20: This is currently the only constructor argument in the spec but doing it this way isn't very future-proof

```suggestion
	public initChannel(...args: ConstructorParameters<typeof BroadcastChannel>): void {
```

Screenshot 2024-10-17 at 09 28 27

libs/utils/src/lib/services/broadcast-channel/broadcast-channel.md
```diff
@@ -0,0 +1,156 @@
...
+
+This `BroadcastChannelService` service wraps around the [BroadcastChannel API](https://developer.mozilla.org/en-US/docs/Web/API/Broadcast_Channel_API) and provides some handy functionality on top of some safety measures. It take SSR into account and will only create channels while in the browser.
```
WHeirstrate, line 3: ... it **takes** SSR into account ...

ian-emsens-shd (changes requested):
Service does not implement the whole spec and lacks automatic type inference. Comment about constructor argument signature also applies to function parameters to a lesser extent.

libs/utils/src/lib/services/broadcast-channel/broadcast-channel.service.ts
WHeirstrate: I'd suggest `NgxUtils: A BroadcastChannel was tried to be initialised without a name. No channel has been initialised, please provide a name for this specific channel.`

libs/utils/src/lib/services/broadcast-channel/broadcast-channel.service.ts
DenisValcke: I've repeated the check for every method to avoid the code from breaking the host application since this will presumably be used for a non-essential features. The developer will be warned of something going wrong, but the flow of the application will not break.

I'll see if I can simplify the checks and streamline the messages and log them as errors to give them a higher priority within the flow.

libs/utils/src/lib/services/broadcast-channel/broadcast-channel.service.ts
DenisValcke: I've not used it to keep the service something that can stand by itself, but I could switch to the service as well if there is a preference to do so.

libs/utils/src/lib/services/broadcast-channel/broadcast-channel.service.ts
DenisValcke: The `postMessage` can be called without calling the `initChannel` first. This was my train of tought when repeating these checks. As said in a previous comment, I'll simplify the checks and messages.

libs/utils/src/lib/services/broadcast-channel/broadcast-channel.service.ts
DenisValcke: Good catch!

libs/utils/src/lib/services/broadcast-channel/broadcast-channel.service.ts
DenisValcke: Will do! Missed that event.

DenisValcke:
To recap the feedback into a todo list:

- [ ] Add tags to the `package.json` in line with the other packages
- [ ] Add the service to the global README of the ngx-utils
- [ ] Fix a typo in the service's README: `it takes SSR into account`
- [ ] Simplify the checks to see if the channelName is provided and exists within the Record
- [ ] Fix the `selectChannel` method mismatch with the documentation
- [ ] Convert `console.warn` messages to `console.error` and format the messages to be in line with others
- [ ] OPTIONAL: use the window service to check if the platform is `browser`.

DenisValcke:
All feedback has been processed (see checklist).

ian-emsens-shd (approved):

IbenTesara (approved):
Still one `BroadcastChannelService` instead of `NgxBroadcastChannelService` in the general docs but looks good to me!

DenisValcke:
- [x ] Update the NgxBroadcastChannelService README to reflect recent changes
- [x] Update the NgxUtils README to reflect recent changes

PR #263, opened 2024-10-21
fix(ngx-inform): Correctly implement aria-labelledBy and aria-describedBy

IbenTesara:


libs/inform/src/lib/abstracts/modal/modal.abstract.component.ts
```diff
@@ -27,4 +51,29 @@ export class NgxModalAbstractComponent<ActionType extends NgxModalActionType, Da
...
 	@Output() public close: EventEmitter<void> = new EventEmitter<void>();
+
+	constructor(
+		@Inject(PLATFORM_ID) private platformId: string,
+		private readonly elementRef: ElementRef<HTMLElement>
+	) {}
+
+	public ngAfterViewInit(): void {
+		// Iben: If we are in the browser, check if either of the two accessibility labels are set
+		if (isPlatformBrowser(this.platformId) && (this.ariaLabelledBy || this.ariaDescribedBy)) {
```
WHeirstrate, line 62: If neither the ariaLabelledBy nor the ariaDescribedBy property were passed, this will not throw an error. Is this by design?

IbenTesara: Yes, as there's also the option to pass an `aria-label`.

WHeirstrate (approved):

PR #267, opened 2024-11-06
feat(table-docs): Add table docs and remaining util docs

SHDBramVandenbussche:
Added all documentation and demos for the table package, add remaining util package docs

### DIFF
--- a/apps/docs/src/app/pages/table/implementation/header/ng-doc.page.ts
+++ b/apps/docs/src/app/pages/table/implementation/header/ng-doc.page.ts
@@ -1,0 +1,16 @@
+import { NgDocPage } from '@ng-doc/core';
+import TableImplementationCategory from 'apps/docs/src/app/categories/table/sub-categories/implementation/ng-doc.category';
+import { TableOneDemoComponent, TableTwoDemoComponent } from './demos';
+
+const HeaderPage: NgDocPage = {
+	title: `Header`,
+	mdFile: './index.md',
+	category: TableImplementationCategory,
+	order: 4,
+	demos: {
+		TableOneDemoComponent,
+		TableTwoDemoComponent,
+	},
+};
+
+export default HeaderPage;
--- a/apps/docs/src/app/pages/table/implementation/header/ng-doc.page.ts
+++ b/apps/docs/src/app/pages/table/implementation/header/ng-doc.page.ts
@@ -1,5 +1,5 @@
 import { NgDocPage } from '@ng-doc/core';
-import TableImplementationCategory from 'apps/docs/src/app/categories/table/sub-categories/implementation/ng-doc.category';
+import { TableImplementationCategory } from '../../../../categories';
 import { TableOneDemoComponent, TableTwoDemoComponent } from './demos';
 
 const HeaderPage: NgDocPage = {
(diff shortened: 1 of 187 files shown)


apps/docs/src/app/pages/table/implementation/header/ng-doc.page.ts
```diff
@@ -0,0 +1,16 @@
+import { NgDocPage } from '@ng-doc/core';
+import TableImplementationCategory from 'apps/docs/src/app/categories/table/sub-categories/implementation/ng-doc.category';
```
DenisValcke, line 2: TODO: make these paths relative instead of absolute from the root of the project.

Perhaps add a barrel file in `apps/docs/src/app/categories` so that the import could be:
```
import { TableImplementationCategory } from '../../../../categories';
```

DenisValcke: This comment also applies to other places where this occurs.

apps/docs/src/app/pages/table/implementation/sorting-selecting-row-actions/demos/table-1/table-one.demo.component.ts
```diff
@@ -0,0 +1,48 @@
...
+	templateUrl: './table-one.demo.component.html',
+})
+export class TableOneDemoComponent {
+	currentSorting: any = undefined;
+	sortableData = [
+		{ name: 'World', firstName: 'Hello', active: true },
+		{ name: 'Tools', firstName: 'NGX', active: false },
+	];
+
+	sort(sortEvent: NgxTableSortEvent) {
```
DenisValcke, line 18: This is something we need to fix on project level, because it has occured a lot in PR's we've opened in the past, but try to always add a modifiers (`private`, `public`, `protected`) and return types to the properties and methods of your classes.

For example here:
```
	public currentSorting: any = undefined;
	public sortableData = [
		{ name: 'World', firstName: 'Hello', active: true },
		{ name: 'Tools', firstName: 'NGX', active: false },
	];
```
☝️ also try to provide specific types for the `currentSorting` and `sortableData` if you can.

And for the method (as an example):
```
	public sort(sortEvent: NgxTableSortEvent): void {...}
```
In this case the method doesn't return anything which is why you can type it with `: void`. Doint this implicitly, makes sure that in the future it is known that this method shouldn't return a value.

DenisValcke: This comment also applies to other classes. I just picked this one as an example.

apps/docs/src/app/pages/utils/pipes/has-observers/demos/has-observers-component/has-observers.demo.component.ts
```diff
@@ -0,0 +1,12 @@
...
+import { HasObserversPipe } from '@ngx/utils';
+
+@Component({
+	standalone: true,
+	imports: [HasObserversPipe],
+	templateUrl: './has-observers.demo.component.html',
+	selector: 'has-observers-demo-component',
+})
+export class HasObserversPipeDemoComponent {
+	@Output() public somethingHasCompleted: EventEmitter<boolean> = new EventEmitter<boolean>();
```
DenisValcke, line 11: Just fyi and not obrigatory, but properties like these can be `readonly` because the can't and shouldn't be overwritten at any point.

DenisValcke:
The feedback I have for this PR can be boiled down to the comments above. Try to take these and apply them to wherever needed and then this PR should be good to go as far as I can tell.

Good work.

PR #268, opened 2024-11-07
feat(home-page): Add home page and routing, move folders

SHDBramVandenbussche:


apps/docs/src/app/pages/docs/landing/docs-page.component.html
```diff
@@ -0,0 +1,7 @@
+<ng-doc-root>
+	<ng-doc-navbar>
+		<h3 style="margin: 0" ngDocNavbarLeft>MyAwesomeDoc</h3>
```
DenisValcke, line 3: I would do either of two things:
1. Add a class and move this `style` to a css/scss file for the component (or use a utility class or something);
2. Or add a TODO comment that makes sure you don't forget to remove this when styling the application.

The danger here is that this attribute is going to get forgetten and cause problems later on when styling the page.

apps/docs/src/app/pages/landing/landing-page.component.ts
```diff
@@ -0,0 +1,16 @@
...
+	standalone: true,
+	imports: [RouterLink, RouterOutlet],
+	selector: 'landing-page',
+	templateUrl: 'landing-page.component.html',
+})
+export class LandingPageComponent implements OnInit {
+	public routes: typeof ERoutes = ERoutes;
+	constructor() {}
+
+	ngOnInit() {}
```
DenisValcke, line 15: The `ngOnInit` and `constructor` can go if not used. We can add them again if needed in the future.

Do a quick check of the codebase to see if this applies to other places as well.

You can always try to enable the [no-useless-constructor](https://eslint.org/docs/latest/rules/no-useless-constructor) and [no-empty-lifecycle-method](https://github.com/angular-eslint/angular-eslint/blob/main/packages/eslint-plugin/docs/rules/no-empty-lifecycle-method.md) ESLint rules to avoid this from happening.

DenisValcke:
FYI: I only checked the files that had changes and skipped over the ones that were just moved.

IbenTesara (approved):

PR #269, opened 2024-11-07
feat(form-docs): Add forms documentation

SHDBramVandenbussche:


apps/docs/src/app/pages/docs/forms/implementation/validators/demos/at-least-one-required/at-least-one-required.demo.component.html
```diff
@@ -0,0 +1,23 @@
...
+	</div>
+
+	<div>
+		<label for="email">
+			Email
+			<input type="email" formControlName="email" name="email" />
+		</label>
+	</div>
+
+	<div *ngIf="form.get('email').errors">
```
IbenTesara, line 16: Kan je kijken om enkel de nieuwe control flow te gebruiken ipv ngIf?

https://angular.dev/guide/templates/control-flow

IbenTesara (changes requested):
Looks good to me already, but using the new control flow is a bit of a must for me.

IbenTesara (approved):

PR #270, opened 2024-11-14
feat(ngx-layout): WCAG/ARIA for drag-and-drop

IbenTesara:


IbenTesara:
@ekincia @WHeirstrate @DenisValcke 

Once we all given our okay on the default text, we'll look into adding more languages 👍

libs/layout/src/lib/directives/drag-and-drop/drag-and-drop-item.directive.ts
```diff
@@ -0,0 +1,258 @@
...
+ */
+@Directive({
+	selector: '[ngxAccessibleDragAndDropItem]',
+	exportAs: 'ngxAccessibleDragAndDropItem',
+	standalone: true,
+	host: {
+		'[attr.tabIndex]': 'tabIndex()',
+	},
+})
+export class NgxAccessibleDragAndDropItemDirective extends NgxHasFocusDragAndDropDirective {
```
WHeirstrate, line 31: I'm gonna be honest, I don't really like the fact that `Accessible` is always visible in the services, directives and types. Yes, we focus on the a11y, but should it be this explicitly mentioned in each name?

WHeirstrate (approved):

libs/layout/src/lib/directives/drag-and-drop/drag-and-drop-item.directive.ts
IbenTesara: I mean, the sole purpose of these directives, types and services is accessibility... They serve no other purpose than that; I don't understand why we'd not include it in the name? 😅 

It's not like we do this with everything we do, this is just 100% intended for accessible drag and drop, I think the intention should be clear here.

libs/layout/src/lib/abstracts/drag-and-drop/drag-and-drop.service.ts
```diff
@@ -0,0 +1,157 @@
...
+	public setDragAndDropDescription(parent: HTMLElement, description?: string): Observable<void> {
+		// Iben: Create the description element and its id
+		const element: HTMLParagraphElement = document.createElement('p');
+		const id: string = UUID.UUID();
+
+		// Iben: Take the current language to fetch the message
+		return (
+			typeof this.currentLanguage === 'string'
+				? of(this.currentLanguage)
+				: this.currentLanguage
```
DenisValcke, line 121: What if a non-string value is passed that is not evaluated as falsy?

libs/layout/src/lib/abstracts/drag-and-drop/drag-and-drop.service.ts
```diff
@@ -0,0 +1,157 @@
...
+	 * @param description - An optional description used to overwrite the default description
+	 */
+	public setDragAndDropDescription(parent: HTMLElement, description?: string): Observable<void> {
+		// Iben: Create the description element and its id
+		const element: HTMLParagraphElement = document.createElement('p');
+		const id: string = UUID.UUID();
+
+		// Iben: Take the current language to fetch the message
+		return (
+			typeof this.currentLanguage === 'string'
```
DenisValcke, line 119: What if this evaluates as a non-string value, but the value is not an observable (for example `undefined` or something else)?

libs/layout/src/lib/components/configurable-layout/configurable-layout.component.ts
```diff
@@ -341,6 +364,32 @@ export class NgxConfigurableLayoutComponent
...
+		const startArray = formData[event.previousContainer];
+		const endArray = formData[event.newContainer];
+
+		// Iben: If the drag and drop is within the same row, we move
+		if (event.previousContainer === event.newContainer) {
+			moveItemInArray(endArray, event.previousIndex, event.newIndex);
+		} else {
+			// Iben: If the drag and drop is over multiple rows, we transfer
+			transferArrayItem(startArray, endArray, event.previousIndex, event.newIndex);
+			formData[event.previousContainer] = startArray;
```
DenisValcke, line 384: Just checking: Is the `previousContainer` always defined?

libs/layout/src/lib/directives/drag-and-drop/drag-and-drop-host.directive.ts
```diff
@@ -0,0 +1,62 @@
...
+	/**
+	 * Returns the container based on the provided index
+	 *
+	 * @param index - The index of the container
+	 */
+	public getContainer(index: number): NgxAccessibleDragAndDropContainerDirective {
+		return this.containers.find((container) => container.index === index);
+	}
+
+	ngAfterViewInit() {
```
DenisValcke, line 56: Member visibility & return type are missing here.

libs/layout/src/lib/directives/drag-and-drop/drag-and-drop-item.directive.ts
DenisValcke, line 1: Check member visibility of the properties. some are missing.

libs/layout/src/lib/directives/drag-and-drop/has-focus.directive.ts
DenisValcke, line 1: Same not here as the previous one.

DenisValcke, line 1: This directive is an abstract, but the classname does not reflect that. Is this intentional?

libs/layout/src/lib/services/live-region/live-region.service.ts
```diff
@@ -0,0 +1,71 @@
...
+		this.region.setAttribute('aria-live', 'assertive');
+		this.region.setAttribute('role', 'log');
+		this.region.setAttribute('aria-relevant', 'additions');
+		this.region.setAttribute('aria-atomic', 'true');
+
+		//Iben: Move the live region out of sight for without assistive technologies
+		hideElement(this.region);
+
+		// Iben: Append the live region to the body
+		document.body.appendChild(this.region);
```
DenisValcke, line 69: Same here.

libs/layout/src/lib/components/configurable-layout/configurable-layout.component.html
```diff
@@ -1,55 +1,63 @@
...
+		[ngxAccessibleDragAndDropContainerIndex]="index"
+		[ngxAccessibleDragAndDropContainerLabel]="rowLabel ? rowLabel + ' ' + (index + 1) : ''"
+		[id]="'ngx-layout-row-' + index"
+		[ngStyle]="{ keys: form.value, showInactive } | ngxConfigurableLayoutItemSize : itemSize"
+		[style.column-gap]="columnGap"
+		[cdkDropListSortPredicate]="beforeSort.bind(this)"
+		[cdkDropListEnterPredicate]="beforeDrop.bind(this)"
+		(cdkDropListDropped)="drop($event)"
+	>
+		@for (item of row; track item; let index = $index) { @if (itemTemplateRecord()[item.key]) {
```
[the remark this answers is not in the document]
IbenTesara: Seems like a thing we need to tackle the moment we bring in all the other packages?

libs/layout/src/lib/abstracts/drag-and-drop/drag-and-drop.service.ts
IbenTesara: I get what you mean here, and I agree on checking for null/undefined; but the typing will already prevent you from returning anything but a string or an Observable. If you decide to bypass this by casting to any then that's on you 😂 

`abstract get currentLanguage(): string | Observable<string>;`

libs/layout/src/lib/components/configurable-layout/configurable-layout.component.ts
IbenTesara: It should be, but I'll double check here as well

libs/layout/src/lib/directives/drag-and-drop/drag-and-drop-host.directive.ts
IbenTesara: Is there an EsLint rule for that somewhere? Might be good to add it to the list ;P

libs/layout/src/lib/const/drag-and-drop/drag-and-drop-message.const.ts
```diff
@@ -0,0 +1,106 @@
...
+		reordered: '{{#item}} wurde von Position {{#from}} nach Position {{#to}} verschoben.',
+		moved: '{{#item}} wurde von {{#from}} nach {{#to}} verschoben.',
+		cancelled: 'Das Verschieben von {{#item}} wurde abgebrochen.',
+		item: 'Eintrag',
+		container: 'Liste',
+
+		description:
+			'Um Einträge in diesem Teil zu verschieben, navigiere mit der Tab-Taste zu einem Eintrag. Mit Drücken der Eingabe- oder Leerzeichentaste wählst du einen Eintrag aus. Mit den Pfeiltasten verschiebst du den Eintrag.',
+	},
+	ku: {
```
DenisValcke, line 85: Just a reminder not to merge this until @ekincia validates these translations.

libs/layout/src/lib/const/drag-and-drop/drag-and-drop-message.const.ts
```diff
@@ -0,0 +1,106 @@
...
+		deselected: '{{#item}} hate jêbirin.',
+		reordered: '{{#item}} ji cîhê {{#from}} ve ji cîhê {{#to}} ve veguhast.',
+		moved: '{{#item}} hate veguhastin ji {{#from}} ve {{#to}}.',
+		cancelled: 'Veguhastina {{#item}} hate betal kirin.',
+		item: 'şanî',
+		container: 'lista',
+		description:
+			'Ji bo veguhastina şanîyan di vê beşê de, bi bikaranîna tuşeya Tab ve her şanîyekê biguherîne. Ji bo hilbijartina şanîyekê, bişkojka Enter an Space bikirtînin. Bi bikaranîna tuşeya şilî şanîyê biguherînin.',
+	},
+	ru: {
```
DenisValcke, line 96: Just a reminder not to merge this until @SergueySHD validates these translations.

PR #272, opened 2024-11-15
feat(layout-docs): add layout docs

SHDBramVandenbussche:


apps/docs/src/app/categories/index.ts
```diff
@@ -4,6 +4,7 @@ import FormsImplementationCategory from './forms/sub-categories/implementation/n
 import I18nCategory from './i18n/ng-doc.category';
```
IbenTesara, line 4: This isn't necessary in this PR, but I'd suggest to start looking into some index files so you can properly export these instead of always directly importing these from the actual files.

You can check how we do that in the libs to see how you can do that, that way it's going to be much easier to read.

IbenTesara (approved):

PR #279, opened 2024-11-29
fix(ngx-layout): Fix a circular dependency issue in webpack builds

IbenTesara:


libs/layout/src/lib/directives/drag-and-drop/drag-and-drop-item.directive.ts
```diff
@@ -152,6 +154,13 @@ export class NgxAccessibleDragAndDropItemDirective extends NgxHasFocusDragAndDro
 			.subscribe();
 	}
 
+	/**
+	 * Registers the item with the drop host
+	 */
+	ngOnInit(): void {
```
DenisValcke, line 160: public

DenisValcke (approved):

PR #303, opened 2024-12-02
feat(global): Move RXJS-utils to repository

IbenTesara:


WHeirstrate (approved):

libs/rxjs/utils/vite.config.ts
DenisValcke, line 1: I see that the vite static copy plugin has been removed from the config. This is needed to include the `docs` folder in the bundle to make sure the links to the per-feature README files are available in npm/github.

project.json
```diff
@@ -0,0 +1,14 @@
+{
+	"name": "workspace",
+	"$schema": "node_modules/nx/schemas/project-schema.json",
+	"targets": {
+		"local-registry": {
+			"executor": "@nx/js:verdaccio",
```
DenisValcke, line 6: So the local registry uses verdaccio? That's pretty clever tbh.

libs/rxjs/utils/vite.config.ts
```diff
@@ -0,0 +1,44 @@
...
+			transformMixedEsModules: true,
+		},
+		lib: {
+			// Could also be a dictionary or array of multiple entry points.
+			entry: 'src/index.ts',
+			name: 'rxjs-utils',
+			fileName: 'index',
+			// Change this to the formats you want to support.
+			// Don't forget to update your package.json as well.
+			formats: ['es', 'cjs'],
```
DenisValcke, line 37: I think `umd` is missing here.

DenisValcke (approved):

project.json
IbenTesara: All Nx here, so no idea what the inner workings are but glad you're happy :D

PR #322, opened 2024-12-10
fix: wrong default date format

reunefe:
Default date formats of date validators were invalid formats. Corrected them so the validators work again when not providing a format


IbenTesara (approved):
Good catch! 🙌

PR #343, opened 2024-12-16
build(ngx-core):v19.0.0 

IbenTesara:
**Description**
Release v19.0.0

**Requirements**

### DIFF
--- a/libs/angular/core/package.json
+++ b/libs/angular/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@studiohyperdrive/ngx-core",
-  "version": "0.0.0",
+  "version": "19.0.0",
   "description": "A library of core functionality used by the Studio Hyperdrive ngx-tools.",
   "keywords": [
     "angular",


libs/angular/core/README.md
WHeirstrate, line 1: 1. Should this markdown file exist?
2. If so, should this not add a little more information as to the purpose of these Angular libs? Perhaps a link to the Angular documentation in a setup section?

libs/angular/core/package.json
```diff
@@ -0,0 +1,21 @@
+{
+	"name": "@studiohyperdrive/ngx-core",
+	"version": "19.0.0",
+  "description": "A library of core functionality used by the Studio Hyperdrive NGX packages.",
+  "keywords": [
+    "angular"
```
WHeirstrate, line 6: Perhaps `Studio Hyperdrive`, `StudioHyperdrive`, and `ngx-tools` could be additional keywords?

WHeirstrate (changes requested):
Shouldn't the target branch be [feat/ngx-core/149-setup-shared-library](https://github.com/studiohyperdrive/ngx-tools/tree/feat/ngx-core/149-setup-shared-library) instead of [master](https://github.com/studiohyperdrive/ngx-tools/tree/master)? Now it looks like you try to both release and implement changes.

libs/angular/core/package.json
WHeirstrate: Additional thought: maybe `ngx-tools` can be a keyword of every package as to connect them more easily through npm?

WHeirstrate (approved):

PR #346, opened 2024-12-31
fix(nx-core): add stylelint

WouterBas:
**Description**
Added nx-stylelint, also upgrade prettier to v3


DenisValcke:
@WouterBas lgtm. The only thing that we need to keep in mind is that it's best to use Prettier for formatting and Stylelint/ESLint for functional linting of code.

PR #354, opened 2025-01-07
Chore/documentation tour demo

reunefe:
**Description**
Moved the tour demo to an entire standalone route.

Normally the demos work just fine, but for the tour demo there is a bit of styling to be done in order to get the scrolling to work correctly. NgDoc however interferes with the tour since both use the CDK overlay, causing the tour to not scroll correctly due to multiple scrollbars. Moving to a separate route for full screen demos like this fixes this completely.

**Requirements**

**Attachments**


IbenTesara:
afbeelding
I did get this error, but other than that it works so I'll already approve this!

IbenTesara (approved):

PR #365, opened 2025-01-09
feature: Add search string util + documentation

reunefe:
**Description**
Adding the normalize, searchValueForSome and searchValueForEvery util + documentation

**Requirements**

**Attachments**


libs/javascript/utils/src/lib/string/search-string/search-string.util.ts
```diff
@@ -0,0 +1,57 @@
+import { normalizeString } from '../normalize-string/normalize-string.util';
+
+export const searchValueForSome = (
```
IbenTesara, line 3: Can you add some TSDoc to this?

IbenTesara (changes requested):

libs/javascript/utils/src/lib/string/search-string/search-string.util.ts
reunefe: Done! Will keep this in mind for the future 😄

IbenTesara (approved):

PR #366, opened 2025-01-09
feature: Add search recursively util + documentation

reunefe:
**Description**
Add search recursively util to javascript utils

**Requirements**

**Attachments**

### DIFF
--- a/libs/javascript/utils/src/lib/object/search-recursively/search-recursively.util.ts
+++ b/libs/javascript/utils/src/lib/object/search-recursively/search-recursively.util.ts
@@ -1,3 +1,9 @@
+/**
+ * Searches recursively over a nested list of items.
+ *
+ * @param list - The list of items we want to loop over recursively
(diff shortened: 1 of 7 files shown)


libs/javascript/utils/src/lib/object/search-recursively/search-recursively.util.ts
```diff
@@ -0,0 +1,18 @@
+export const searchRecursively = <ItemType extends { children: ItemType[] }>(
```
IbenTesara, line 1: Can you add some documentation (inline and TSDoc)?

IbenTesara (changes requested):

libs/javascript/utils/src/lib/object/search-recursively/search-recursively.util.ts
reunefe: Done! Will keep this in mind for the future 😄

PR #372, opened 2025-01-13
documentation(ngx-table): Update introduction page

IbenTesara:
**Description**
Updated the introduction page to better reflect our goals. Updated the styling slightly so we can have non-dev playgrounds

**Requirements**

**Attachments**
afbeelding


apps/docs/src/app/pages/docs/angular/table/introduction/demos/configuration/table-configuration.demo.component.ts
```diff
@@ -0,0 +1,40 @@
...
+	public readonly currentSorting: WritableSignal<NgxTableSortEvent> = signal({
+		column: 'country',
+		direction: NgxTableSortDirection.ASCENDING,
+	});
+	public control: FormControl<any> = new FormControl<any>([]);
+
+	@Input() public Selectable: boolean = true;
+	@Input() public Sortable: boolean = true;
+	@Input() public Columns: 'All' | 'Minimal' = 'All';
+	@Input() public Detail: boolean = true;
```
WHeirstrate, line 28: Is there a reason these inputs and the string value of `Columns` are capitalised?

apps/docs/src/app/pages/docs/angular/table/introduction/demos/styling/table-styling.demo.component.ts
```diff
@@ -0,0 +1,21 @@
...
+	imports: [NgxTable, CommonModule, ReactiveFormsModule],
+	selector: 'table-styling-demo',
+	templateUrl: './table-styling.demo.component.html',
+	styleUrl: './table-styling.demo.component.scss',
+	host: {
+		class: 'no-controls settings-on-top',
+	},
+})
+export class NgxTableStylingDemoComponent {
+	@Input() public Styling: 'Simple' | 'Modern' | 'Classic' | 'Hyperdrive' | undefined;
```
WHeirstrate, line 18: Same as before: what's the reason this key and its values are capitalised? Capitalised classes look very strange, is this a requirement for the `ng-doc` package? The same comment applies to this component's stylesheet.

apps/docs/src/app/pages/docs/angular/table/introduction/index.md
```diff
@@ -1,19 +1,35 @@
...
+`ngx-table` is an Angular CDK based package to quickly create complex data tables.
 
-## Angular Tools: NgxTable (`@studiohyperdrive/ngx-table`)
+## For organisations
 
-This library provides a quick and easy template based table builder using the Angular CDK Table.
+In many applications, data-tables are the primary way to both visualize and interact with complex datasets. Our table package aims to be a robust all-in-one solution to handle these data-tables in a consistent way, with a focus on user-experience and accessibility.
 
-## Concept
+Designed to be catered to your individual needs, the data-table comes with little to no styling up front. Using this approach, there's less restrictions for developers to adjust the styling to your visual identity. In the example below you can switch the styling of the presented table to see different approach to the visuals.
```
WHeirstrate, line 10: Semantics remark: `Designed to be catered to your individual needs, ...` seems to convey the message that we have a product that is already providing the company (or devs) with whatever they need. We are not `catered`, we are _`caterable`_ - though I doubt that's a proper word. 😅

apps/docs/src/app/pages/docs/angular/table/introduction/index.md
```diff
@@ -1,19 +1,35 @@
...
+
+{{ NgDocActions.playground("TableConfigurationPlayground") }}
+
+With a focus on reusability, our solution comes with the ability to define cells that can be used across the entire application. If certain data, like an identification number, always has to be formatted in a specific way then the package allows for the creation of a cell specifically designed for that data. Not only does this guarantee consistency for your end-users, changes to these cells can be done quickly and safely without worries.
+
+Data-tables often visualize large and complex datasets. To ensure that every end-user is able to interact and interpret the provided data, our data-table is fully WCAG/WAI-ARIA compliant. Fully navigable using a keyboard, this means that the table and its data is always accessible for all users, regardless of disabilities.
+
+Looking for more information on how this package can help you out in your application? Mail us at [info@studiohyperdrive.be](mailto:info@studiohyperdrive.be)
```
WHeirstrate, line 22: Is a DL a cleaner solution? I doubt anyone would mail us, but regardless...

apps/docs/src/app/pages/docs/angular/table/introduction/index.md
```diff
@@ -1,19 +1,35 @@
...
+
+## For developers
+
+The `ngx-table` package is based on the existing Angular CDK table. Whilst the existing package is a very useful tool to build tables with, it is known to have a rather complex syntax and does not feature several common use-cases out of the box.
+
+Using a template based system, the package provides a quick and easy way to build large complex data tables with, including several common use-cases such as selecting one or multiple rows, column based sorting, detail rows, actions, etc.
+
+Whilst the package itself opinionated and is tailored to be WCAG/WAI-ARIA compliant by default, the template based approach offers a lot of room to customize the table to the required use-case.
+
+Looking for more information on how this package can help you out in your application? Check out the `Implementation` link on the left to find out more!
```
WHeirstrate, line 32: Nitpicking, but `on the left` doesn't apply to mobile views. There, the navigation is on the right.

WHeirstrate (changes requested):

apps/docs/src/app/pages/docs/angular/table/introduction/demos/styling/table-styling.demo.component.ts
IbenTesara: I'm aware that it looks strange; but the playground uses these inputs directly as is.

We can either:

- Write custom controls for each input
- Use custom styling to overwrite specific inputs
- Capitalize the inputs

In my honest opinion I think it's much easier to maintain this by just capitalizing the inputs and their options than to strictly adhere to Angular styleguides here. It looks strange I agree, but it's the most pragmatic solution for something that isn't going to be used anywhere but in the docs.

WHeirstrate: I agree. Perhaps an inline comment could prevent future devs from breaking the styling by informing them of this approach.

apps/docs/src/app/pages/docs/angular/table/introduction/index.md
IbenTesara: Do you have a suggestion how to rephrase this? I can somewhat understand what you mean that people could interpret it like that; but on the other hand I don't really know if there's any phrasing that's going to completely avoid that without making it super convoluted and longer than it needs to be 😅

apps/docs/src/app/pages/docs/angular/table/introduction/index.md
IbenTesara: Good catch!

apps/docs/src/app/pages/docs/angular/table/introduction/demos/styling/table-styling.demo.component.ts
IbenTesara: I originally had an inline comments for it; but then again we'd have to repeat the same comment everywhere which also seems pretty counterproductive.

I think we should just catch it in PRs instead and see if we run into issues with it. If we do, then we can look into better ways to handle it (eslint rules).

apps/docs/src/app/pages/docs/angular/table/introduction/index.md
WHeirstrate: My suggestion would be to simply replace the word with another:
`Designed to adapt to your individual needs, the data-table comes with little to no styling up front.`
Or maybe something bolder:
`Built to suit your unique requirements, the data-table comes with minimal styling out of the box.`

WHeirstrate (approved):

PR #374, opened 2025-01-14
feature: update highlight pipe to be much more flexible

reunefe:
**Description**
Update the highlightPipe to be much more flexible.

**Requirements**

**Attachments**


libs/angular/utils/src/lib/pipes/highlight/highlight.pipe.ts
```diff
@@ -1,25 +1,129 @@
...
+		if (!value || !highlight || typeof value !== 'string' || typeof highlight !== 'string') {
 			return value;
 		}
 
-		// Iben: Match the highlight and replace it with the wrapped element
-		const regexp = new RegExp(`(${highlight})`, 'i');
-		return value.replace(regexp, `<${domElement}>$1</${domElement}>`);
+		// Femke: Normalize all characters => é â ö etc will be replaced with e a o
+		const normalizedValue = normalizeString(value);
+		const normalizedHighlight = normalizeString(highlight);
```
WHeirstrate, line 44: Even though we can provide `false` to the config object, we will always make the computations for the normalization. I know this is a type of micro-optimalisation, but I'd only normalize the value if the user requested this. Otherwise, this computation was done for naught.

libs/angular/utils/src/lib/pipes/highlight/highlight.pipe.ts
```diff
@@ -1,25 +1,129 @@
...
 	 * @param value - The full text with the part we wish to highlight
 	 * @param highlight - The part of the text we wish to highlight
-	 * @param domElement - The dom-element we wish to use to highlight, by default this is a `mark` element
+	 * @param config - The configuration to determine if we want to normalize the values, to be case-sensitive, which tag and/or class to use for the highlight
+	 * @param config.normalized - search for matches should be normalized. Default = true
+	 * @param config.caseInsensitive - search for matches should be caseInsensitive. Default = true
+	 * @param config.splitTextToHighlight - text to highlight should be split on space. Default = false
+	 * @param config.someOrEveryMatch - highlight only the first match or all matches. Default = 'every'
+	 * @param config.tag - tag used to highlight. Default = 'mark'
+	 * @param config.highlightClass - class used to highlight. Default = 'mark-highlight'
```
WHeirstrate, line 22: I love the TSDoc, but I'd move the definition of the keys of the object to the type definition in `./highlight.types.ts`. That way, the information of the keys will be visible in users' IntelliSense. 🙂 
I'd keep only the default values here.

libs/angular/utils/src/lib/pipes/highlight/highlight.pipe.ts
```diff
@@ -1,25 +1,129 @@
...
+		const regEx = new RegExp(usableHighlight, regexFlags);
+
+		// Femke: Use a custom replacer so we can update our marked indices in case we found a match with our regex
+		let replacedResult = usableValue.replace(regEx, (match, index) => {
+			if (!match) {
+				return '';
+			}
+
+			const indexAsNumber = index as number;
+			const endIndex = indexAsNumber + match.length;
```
WHeirstrate, line 92: Why do we cast explicitly as a number? If the `index` would be `undefined`, the value would be `NaN` after the sum. 🤔

libs/angular/utils/src/lib/pipes/highlight/highlight.pipe.ts
```diff
@@ -1,25 +1,129 @@
...
+			// Femke: we are using the original text here to make sure we have the original accent characters within the highlight
+			const result = `<${tag} class="${highlightClass}">${value.substring(indexAsNumber, indexAsNumber + match.length)}</${tag}>`;
```
WHeirstrate, line 94: If the user provided an empty string, the DOM element would be `<tag class="">value</tag>` which would be DOM pollution with an empty class. 🙂

WHeirstrate (changes requested):

libs/angular/utils/jest.config.ts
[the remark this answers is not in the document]
IbenTesara: Honestly, I don't think it's bad to do it gradually.

That being said, I'm waiting to see what Angular will do with it. [Their recent blogpost](https://blog.angular.dev/angular-2025-strategy-9ca333dfc334) pointed at them still looking at the options.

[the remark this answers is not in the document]
reunefe: For now I removed the configuration for jest again.

libs/angular/utils/src/lib/pipes/highlight/highlight.pipe.ts
reunefe: Probably a leftover from the passed. Changed accordingly

reunefe:
@WHeirstrate changes have been made and updated the docs as well to have a better playground 😉

libs/angular/utils/src/lib/pipes/highlight/highlight.pipe.ts
```diff
@@ -1,25 +1,124 @@
...
+		highlight: string | null,
+		config?: HighlightConfig | null
+	): string {
+		// Femke: Setup configuration or defaults
+		const normalized = config?.normalized ?? true;
+		const caseInsensitive = config?.caseInsensitive ?? true;
+		const splitTextToHighlight = config?.splitTextToHighlight ?? false;
+		const someOrEveryMatch = config?.someOrEveryMatch || 'every';
+		const tag = config?.tag || 'mark';
+		const highlightClass = config?.highlightClass ?? 'mark-highlight';
```
IbenTesara, line 34: We usually start our classes with `ngx-`, so that would become `ngx-mark-highlight`.

libs/angular/utils/src/lib/types/highlight.types.ts
```diff
@@ -0,0 +1,18 @@
...
+ * @property normalized - search for matches should be normalized
+ * @property caseInsensitive - search for matches should be caseInsensitive
+ * @property splitTextToHighlight - text to highlight should be split on space.
+ * If false the entire string will be used for searching e.g. "some value"
+ * If true the value will be splot e.g. "some value"
+ * @property someOrEveryMatch - highlight only the first match or all matches
+ * @property tag - tag used to highlight
+ * @property highlightClass - class used to highlight
+ */
+export interface HighlightConfig {
```
IbenTesara, line 11:
```suggestion
export interface NgxHighlightConfiguration {
```

libs/angular/utils/src/lib/pipes/highlight/highlight.pipe.ts
```diff
@@ -1,25 +1,124 @@
 import { Pipe, PipeTransform } from '@angular/core';
+import { normalizeString } from '@studiohyperdrive/utils';
+
+import { HighlightConfig } from '../../types';
 
 @Pipe({
 	name: 'highlight',
-	standalone: true,
 })
 export class HighlightPipe implements PipeTransform {
```
IbenTesara, line 9:
```suggestion
export class NgxHighlightPipe implements PipeTransform {
```

IbenTesara (changes requested):
Some small changes to make this more inline with the rest of the lib

IbenTesara (approved):

libs/angular/utils/src/lib/types/highlight.types.ts
```diff
@@ -1,4 +1,14 @@
...
+/**
+ * @property normalized - search for matches should be normalized
+ * @property caseInsensitive - search for matches should be caseInsensitive
+ * @property splitTextToHighlight - text to highlight should be split on space.
+ * If false the entire string will be used for searching e.g. "some value"
+ * If true the value will be split e.g. "some" or "value"
+ * @property someOrEveryMatch - highlight only the first match or all matches
+ * @property tag - tag used to highlight
+ * @property highlightClass - class used to highlight
+ */
```
WHeirstrate, line 10: This does not trigger the IntelliSense afaik. To provide inline documentation, the interface should look like this:
```TypeScript
export interface NgxHighlightConfiguration {
	/**
	 * Whether the search for matches in the provided string should be normalized.
	 */
	normalized?: boolean;
	/**
	 * Whether search for matches should be case insensitive.
	 */
	caseInsensitive?: boolean;
	/**
	 * Highlight only the first match or all matches.
	 */
	someOrEveryMatch?: 'some' | 'every';
	/**
	 * The way in which the highlightable text should be split on a `space`.
	 * If `false`, the entire string will be used for searching (e.g. "some value").
	 * If `true`, the value will be split (e.g. "some" or "value")
	 */
	splitTextToHighlight?: boolean;
	/**
	 * The HTML tag used to wrap the highlighted text (snippet) with.
	 */
	tag?: string;
	/**
	 * A custom class that will be applied to every {@link tag}`.
	 */
	highlightClass?: string;
}
```

WHeirstrate (changes requested):

libs/angular/utils/src/lib/types/highlight.types.ts
reunefe: Weird, it triggered on my end. But I'll just add both, just in case

WHeirstrate: Could you try in VSCode instead of WebStorm? If it triggers on your end in VSCode, then you may ignore my remark. 🙂

reunefe: Adding the documentation inline, would not be TS Docs and thus defeats the purpose I think.

In VS code on hover over each property of  I get the information provided, regardless if it is inline or as TS docs.
However on hover over the configuration interface itself I only see the TS docs provided, not the inline ones.

In VS Webstorm on hover over each property of I get the information provided inline, not the TS docs itself
However on hover over the configuration interface itself I only see the TS docs provided, not the inline ones.

So VS Code seems to work a bit better than Webstorm in that regards. But seems to me TS Docs as shown above are the best solution?

reunefe: Changed to be more inline with everything else.
Added a generic for the interface and moved the specifics to each property

WHeirstrate (approved):
👌

PR #378, opened 2025-01-16
feat(ngx-layout): Add image-marker component

IbenTesara:
**Description**
Adds an Angular wrapper to [MarkerJs2 and Marker-Live](https://markerjs.com/).

**Requirements**

**Attachments**
afbeelding
afbeelding


tsconfig.base.json
```diff
@@ -33,7 +33,7 @@
 			"@js/regex": ["libs/javascript/regex/src/index.ts"],
 			"@js/testing": ["libs/javascript/testing/src/index.ts"],
 			"@express/status": ["libs/express/status/src/index.ts"],
-			"@studiohyperdrive/ngx-core": ["libs/angular/core/src/index.ts"]
+			"@ngx/core": ["libs/angular/core/src/index.ts"]
```
IbenTesara, line 36: This needed to be changed, as otherwise we can't depend on the published ngx-core package.

libs/angular/layout/src/lib/components/image-marker/image-marker.component.ts
```diff
@@ -0,0 +1,165 @@
...
+				{
+					mode: this.canEdit ? 'edit' : 'view',
+					allowZoom: true,
+					defaultState: this.startState || undefined,
+					markerTypes: this.markerTypes,
+				}
+			);
+
+			// Iben: Listen to the valueChanges based on the provided type.
+			// Both Observables are closed in the NgxImageMarkerComponent and therefor do not require a takeUntil here
```
reunefe, line 143: This seems a bit confusing to me. It says here they do not require a takeUntil, but still they have one.

libs/angular/layout/src/lib/components/image-marker/image-marker.component.ts
```diff
@@ -0,0 +1,165 @@
...
+		// Iben: Recreate the marker whenever the configuration is adjusted
+		this.createMarker();
+	}
+
+	/**
+	 * Creates a MarkerJs view based on the provided configuration
+	 */
+	private createMarker() {
+		// Iben: Only create the image when we're in the browser
+		this.windowService.runInBrowser(() => {
```
reunefe, line 123: Is this really necessary? The createImageMarker already checks if we are in the browser.

libs/angular/layout/src/lib/services/image-marker/image-marker.service.ts
```diff
@@ -0,0 +1,189 @@
...
+	 * Listen to the value changes in the editable view
+	 *
+	 * @param {MarkerArea} marker - The marker view
+	 */
+	private createMarkerValueChanges(marker: MarkerArea): Observable<MarkerAreaState> {
+		// Iben: Setup valueChanges
+		const valueChanges = new Subject<MarkerAreaState>();
+
+		// Iben: Update the subject whenever a new item was added to the marker view
+		marker.addEventListener('statechange', (event) => {
```
reunefe, line 164: Probably just me being overly cautious, but adding event listeners without them being cleaned up, could cause memory leaks

apps/layout-test/src/pages/image-marker/image-marker.component.ts
```diff
@@ -0,0 +1,30 @@
+import { Component } from '@angular/core';
+import { NgxImageMarkerComponent, NgxImageMarkerState } from '@ngx/layout';
+
+@Component({
+	selector: 'ngx-image-marker-page',
+	templateUrl: './image-marker.component.html',
+	imports: [NgxImageMarkerComponent],
```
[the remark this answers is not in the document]
IbenTesara: Yeah I usually don't bother when it's in the test apps 😅

libs/angular/layout/src/lib/components/image-marker/image-marker.component.ts
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	ngOnChanges(): void {
+		// Iben: If no marker exists or if the image has not rendered, early exit
+		if (!this.currentMarker || !this.imageElement) {
+			return;
+		}
+
+		// Iben: Recreate the marker whenever the configuration is adjusted
+		this.createMarker();
```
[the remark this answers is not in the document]
IbenTesara: Yeah you're right, I'll see about using the simpleChangesUtil! ;)

libs/angular/layout/src/lib/components/image-marker/image-marker.component.ts
IbenTesara: Oh wait, I think I copy pasted that, this shouldn't be here 😂

libs/angular/layout/src/lib/components/image-marker/image-marker.component.ts
IbenTesara: I've done it here so that I don't have to do further if checks on whether the marker exists; but it could be done either way indeed.

libs/angular/layout/src/lib/services/image-marker/image-marker.service.ts
IbenTesara: You're right, I assumed that the package did it automatically because the docs didn't mention it; but in their class list there is a mention of a remover. I'll add that!

apps/layout-test/src/pages/image-marker/image-marker.component.ts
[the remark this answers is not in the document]
reunefe: I can understand, however in for example the configurable layout compoent, we noticed when using it in a component that had OnPush for the change detection, the layout component did not respond as it should be.
So maybe we should start putting it to onPush so in the future we will not run into issues and need to update the library itself because something is not working which we could have detected if our test app would cover it? 🙈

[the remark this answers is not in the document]
IbenTesara: That sounds logical indeed, I'll add it here!

libs/angular/layout/src/lib/components/image-marker/image-marker.component.ts
```diff
@@ -0,0 +1,184 @@
...
+	NgxImageMarkerTypes,
+} from '../../types';
+
+/**
+ * A component wrapper for MarkerJs views
+ *
+ * https://markerjs.com/
+ */
+
+// TODO: Iben: Once we figured out how we'll share the FormAccessors with other packages, we should make this a ControlValueAccessor
```
WHeirstrate, line 31: Perhaps create an issue of this obstacle?

libs/angular/layout/src/lib/components/image-marker/image-marker.component.ts
```diff
@@ -0,0 +1,184 @@
...
+
+	/**
+	 * Whether we can edit the view, by default this is true
+	 */
+	@Input() public canEdit: boolean = true;
+
+	/**
+	 * An optional record of types of Markerjs markers we wish to render
+	 */
+	@Input() public markerTypes: NgxImageMarkerTypes;
```
WHeirstrate, line 85: L62:L85
Why don't we use `signalInput`s in this component?

libs/angular/layout/src/lib/components/image-marker/image-marker.component.ts
```diff
@@ -0,0 +1,184 @@
...
+
+	ngAfterViewInit(): void {
+		// Iben: Create the initial marker
+		this.createMarker();
+	}
+
+	ngOnChanges(changes: SimpleChanges): void {
+		// Iben: If no marker exists or if the image has not rendered, early exit
+		if (!this.currentMarker || !this.imageElement) {
+			return;
```
WHeirstrate, line 113: Does this imply that something was initiated incorrectly? Or will this hook also run before the element could possible be initiated? Or should this component not be initiated on the server, for instance?

libs/angular/layout/src/lib/services/image-marker/image-marker.service.ts
```diff
@@ -0,0 +1,227 @@
...
+@Injectable({
+	providedIn: 'root',
+})
+export class NgxImageMarkerService implements OnDestroy {
+	/**
+	 * A record with all markers
+	 */
+	private readonly markers: Record<string, NgxImageMarker> = {};
+
+	constructor(private readonly windowService: NgxWindowService) {}
```
WHeirstrate, line 31: Do we not allow for a global config provided in the `providers` array in the application-bootstap?

package.json
```diff
@@ -1,140 +1,143 @@
 {
```
WHeirstrate, line 1: Not blocking, just a remark: we really need to fix the linter. This entire file gets changed by every person that makes an edit, which messes with this file's git-blame.

package.json
```diff
@@ -1,140 +1,143 @@
...
-    "ts-jest": "^29.2.5",
-    "ts-node": "10.9.2",
-    "typescript": "5.6.3",
-    "vite": "5.4.11",
-    "vite-plugin-dts": "4.3.0",
-    "vite-plugin-static-copy": "2.2.0",
-    "vitest": "2.1.8"
-  }
+	"name": "workspace",
+	"version": "18.0.0",
```
WHeirstrate, line 3: Not blocking, just a remark: why is this version never updated? Should this not at least reflect the current `@angular/core` version?

libs/angular/layout/src/lib/components/image-marker/image-marker.component.ts
IbenTesara: I don't think we use them anywhere at this point in time, but I'll add a ticket to just run the migration script for it.

libs/angular/layout/src/lib/components/image-marker/image-marker.component.ts
IbenTesara: No, [ngOnChanges will run already before the ngAfterViewInit](https://angular.dev/guide/components/lifecycle#during-initialization); this simply prevents a marker from being created before the ngAfterViewInit has run, where we want the initial one to be created.

libs/angular/layout/src/lib/components/image-marker/image-marker.component.ts
IbenTesara: Yes, I still need to finish up the RFC for that.

package.json
IbenTesara: I know, I thought it was fixed recently but it doesn't seem to have been fixed 🥲

package.json
IbenTesara: Given that this is now an Angular and non-Angular monorepo, this will update the next time we upgrade Nx.

WHeirstrate (approved):

PR #385, opened 2025-01-21
feat(ngx-layout): Allow zoom levels to be passed to the image-marker

IbenTesara:
**Description**
Allows us to pass the current zoom level and the possible zoom levels to the image marker

**Requirements**

**Attachments**
afbeelding

### DIFF
--- a/libs/angular/layout/src/lib/services/image-marker/image-marker.service.ts
+++ b/libs/angular/layout/src/lib/services/image-marker/image-marker.service.ts
@@ -144,6 +144,8 @@ export class NgxImageMarkerService implements OnDestroy {
 		// These can later on be extended when needed
 		marker.uiStyleSettings.zoomButtonVisible = configuration.allowZoom;
 		marker.uiStyleSettings.zoomOutButtonVisible = configuration.allowZoom;
+		marker.zoomSteps = configuration.zoom?.levels || [1, 2, 3, 4];
+		marker.zoomLevel = configuration.zoom?.current ?? 1;
 		marker.uiStyleSettings.clearButtonVisible = configuration.allowClear;
 
 		// Iben: Set the available marker types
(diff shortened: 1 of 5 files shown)


libs/angular/layout/src/lib/services/image-marker/image-marker.service.ts
```diff
@@ -144,6 +144,8 @@ export class NgxImageMarkerService implements OnDestroy {
 		// These can later on be extended when needed
 		marker.uiStyleSettings.zoomButtonVisible = configuration.allowZoom;
 		marker.uiStyleSettings.zoomOutButtonVisible = configuration.allowZoom;
+		marker.zoomSteps = configuration.zoom?.levels || [1, 2, 3, 4];
+		marker.zoomLevel = configuration.zoom?.current || 1;
```
reunefe, line 148: So 0 as zoomLevel is never an option? In that case I would add a remark in the docs

IbenTesara: Good catch, I did it in the other component but not here 🙈

libs/angular/layout/src/lib/services/image-marker/image-marker.service.ts
```diff
@@ -144,6 +144,9 @@ export class NgxImageMarkerService implements OnDestroy {
 		// These can later on be extended when needed
 		marker.uiStyleSettings.zoomButtonVisible = configuration.allowZoom;
 		marker.uiStyleSettings.zoomOutButtonVisible = configuration.allowZoom;
+		marker.zoomSteps = configuration.zoom?.levels || [1, 2, 3, 4];
+		marker.zoomLevel =
+			configuration.zoom?.current === undefined ? 1 : configuration.zoom.current;
```
reunefe, line 149: Why not simply `configuration.zoom?.current ?? 1` Is shorter and easier to read in my opinion.
In case of null/undefined it would fallback to 1 in all other cases it would take the given value, even 0

IbenTesara: I've always found this notation to be clearer in intent to me; but as you said it's shorter so I'll change it :)

reunefe (approved):

PR #389, opened 2025-01-29
feat(js-types-auth): Setup authentication types package

IbenTesara:
**Description**
Sets up a library with shared authentication types that we can later use in shared authentication frontend packages

**Requirements**

### DIFF
--- a/libs/javascript/types/auth/.eslintrc.json
+++ b/libs/javascript/types/auth/.eslintrc.json
@@ -1,0 +1,25 @@
+{
+	"extends": ["../../../../.eslintrc.json"],
+	"ignorePatterns": ["!**/*"],
+	"overrides": [
+		{
+			"files": ["*.ts", "*.tsx", "*.js", "*.jsx"],
+			"rules": {}
+		},
+		{
+			"files": ["*.ts", "*.tsx"],
+			"rules": {}
+		},
+		{
+			"files": ["*.js", "*.jsx"],
+			"rules": {}
+		},
+		{
+			"files": ["*.json"],
+			"parser": "jsonc-eslint-parser",
+			"rules": {
+				"@nx/dependency-checks": "error"
+			}
+		}
+	]
+}
--- a/libs/javascript/types/auth/package.json
+++ b/libs/javascript/types/auth/package.json
@@ -1,0 +1,27 @@
+{
+  "name": "@studiohyperdrive/types-auth",
+  "version": "0.0.1",
+  "dependencies": {},
+  "type": "commonjs",
+  "main": "./index.cjs",
+  "typings": "./index.d.ts",
+  "description": "A package containing common authentication related types",
+  "keywords": [
+    "authentication",
+    "user",
+    "session",
+    "features",
+    "permissions"
+  ],
+  "homepage": "https://open-source.studiohyperdrive.be/docs/javascript/types/introduction",
+  "author": {
+    "name": "Studio Hyperdrive",
+    "url": "https://studiohyperdrive.be/"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/studiohyperdrive/hyperdrive-opensource",
+    "directory": "libs/javascript/types/auth/src"
+  },
+  "license": "MIT"
+}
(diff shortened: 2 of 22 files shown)


jsprds:
@IbenTesara can you give more information about the Verdaccio setup? 🙏

package.json
```diff
@@ -136,9 +138,13 @@
...
 		"ts-node": "10.9.2",
 		"typescript": "5.6.3",
+		"verdaccio": "^5.0.4",
 		"vite": "5.4.11",
 		"vite-plugin-dts": "4.3.0",
 		"vite-plugin-static-copy": "2.2.0",
 		"vitest": "2.1.8"
+	},
+	"nx": {
+		"includedScripts": []
```
WHeirstrate, line 148: Why was this added?

package.json
```diff
@@ -48,8 +48,8 @@
 		"@nx/eslint-plugin": "20.2.1",
 		"@nx/vite": "20.2.2",
 		"@studiohyperdrive/env": "1.0.4",
-		"@studiohyperdrive/utils": "^2.0.0",
 		"@studiohyperdrive/ngx-core": "19.1.0",
+		"@studiohyperdrive/utils": "^2.0.0",
```
WHeirstrate, line 52: Why don't we use a fixed versioning here?

IbenTesara:
> @IbenTesara can you give more information about the Verdaccio setup? 🙏

It's something that was auto-generated by Nx, we used it in the previous repos as well but I don't have that much more insight in it I'm afraid 😅

libs/javascript/types/auth/.eslintrc.json
```diff
@@ -0,0 +1,33 @@
...
+				"@nx/dependency-checks": [
+					"error",
+					{
+						"ignoredFiles": [
+							"{projectRoot}/eslint.config.{js,cjs,mjs}",
+							"{projectRoot}/esbuild.config.{js,ts,mjs,mts}"
+						]
+					}
+				]
+			}
```
WHeirstrate, line 30: Why does this divert from the global `.eslintrc.json`? The global config lacks this rule, so the question is actually why this rule should only be enforced in this lib.

WHeirstrate (changes requested):
Could you illustrate what purpose this authentication package serves? I assume this is part 1 of many to create a plug-and-play authentication service that communicates between a node backend and a frontend? Will this be framework agnostic, or catered to Angular?

package.json
IbenTesara: Automatically added by Nx.

bertyhell (approved):

package.json
IbenTesara: This isn't part of the scope of this PR nor was it added by me; but I see nothing wrong with this not being a fixed version.

We're responsible for that package, ergo we should take responsibility to ensure that we can keep working with this approach instead of setting it on a fixed version and updating it with every minor release we do for the utils package. 🤷

IbenTesara:
> Could you illustrate what purpose this authentication package serves? I assume this is part 1 of many to create a plug-and-play authentication service that communicates between a node backend and a frontend? Will this be framework agnostic, or catered to Angular?

Honestly I don't see how much more information I can give than the one in the docs 😅 

This is indeed a framework agnostic types package, so both the frontend packages and the backend packages or backends that want to link with the frontend packages can use these types. 

The end goal is to have a frontend package for each framework, making it so that we can easily handle permissions, feature flags, and user management in a consistent way :)

package.json
WHeirstrate: I missed the fact that you moved this line down. I though you had added this, my bad.

libs/javascript/types/auth/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "@studiohyperdrive/types-auth",
+  "version": "0.0.1",
```
WHeirstrate, line 3: This version does not match the one in the release. Is this a conscious decision?

WHeirstrate (approved):

libs/javascript/types/auth/package.json
WHeirstrate: Nevermind, I presume you want to change this in the build PR?

IbenTesara: Correct :D

PR #400, opened 2025-02-06
feat(ngx-auth): Add interceptor and authenticated http client

IbenTesara:
**Description**
Adds the `NgxAuthenticatedHttpClient` to the ngx-auth package.

**Requirements**

**Attachments**
afbeelding

### DIFF
--- a/libs/angular/authentication/src/lib/services/authenticated-http-client/authenticated-http-client.service.ts
+++ b/libs/angular/authentication/src/lib/services/authenticated-http-client/authenticated-http-client.service.ts
@@ -1,0 +1,170 @@
+import { HttpClient, HttpContext, HttpResponse } from '@angular/common/http';
+import { Inject, Injectable } from '@angular/core';
+import * as clean from 'obj-clean';
+
+import { map, Observable } from 'rxjs';
+import { NgxAuthenticatedHttpClientConfiguration } from '../../types';
+import { NgxAuthenticationUrlHandlerToken } from '../../tokens';
+
+/**
+ * An opinionated wrapper of the HttpClient providing easy ways to make authenticated and unauthenticated calls
+ */
+@Injectable({ providedIn: 'root' })
+export class NgxAuthenticatedHttpClient {
+	private baseUrl: string;
+
+	constructor(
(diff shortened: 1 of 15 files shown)


libs/angular/authentication/src/lib/services/authenticated-http-client/authenticated-http-client.service.ts
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	/**
+	 * Constructs a GET request to the provided API
+	 *
+	 * @param  url - The url of the API
+	 * @param params - An optional set of params we wish to send to the API
+	 * @param withCredentials - Whether the call is made by an authenticated user, by default true
+	 */
+	public get<DataType = unknown>(
```
DenisValcke, line 40: The only thing that I want to note is that, for our project, we've added the ability to provide `HttpContext` in these functions.

For example:
```
	public get<T>(url: string, params?: any, withCredentials: boolean = true, context?: HttpContext): Observable<T> {
		return this.httpClient.get<T>(this.composeUrl(url), {
			params,
			withCredentials,
			...(context ? { context } : {})
		});
	}
```

We, for example, use this to let an interceptor know to skip global error handling for certain requests. But this can be used for a lot of other cases as well.

Could be a good addition here to make the methods more versatile.

IbenTesara: Sounds good, I'll add it in!

DenisValcke: For your documentation. This could be a potential use case:

DenisValcke: For your documentation. This could be a potential use case:

In your component or service, add HttpContext like this:
```typescript
const context = new HttpContext().set(DO_SOMETHING, true)

return service.get(url, params, withCredentials, context);
```

In an interceptor you can now get that context:
```typescript
const doSomething: boolean = request.context.get(DO_SOMETHING);

if (doSomething) {
  doSomething();
}
```

PR #409, opened 2025-02-09
build(store): v20.0.0

ekincia:
**Description**
Release ngx-store v20.0.0

**Requirements**

### DIFF
--- a/libs/angular/store/package.json
+++ b/libs/angular/store/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "@studiohyperdrive/ngx-store",
-	"version": "19.0.0",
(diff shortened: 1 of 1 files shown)


libs/angular/store/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@studiohyperdrive/ngx-store",
-	"version": "19.0.0",
+	"version": "20.0.0",
```
IbenTesara, line 3: This version should be `19.1.0`, because the major version will always follow the current major Angular version.

IbenTesara (changes requested):

PR #413, opened 2025-03-17
fix(global): Replace Angular-uuid with uuid

IbenTesara:
**Description**
Changed angular-uuid to uuid. Some files had incorrect formatting which also have been fixed.

**Requirements**


SergueySHD (approved):

DenisValcke (approved):
Merci om dit al op te nemen!

PR #431, opened 2025-05-21
Fix: correct next page logic in calculateNextPage utility

Krishprajapati15:
Fixed an issue in `calculateNextPage` where it incorrectly allowed next page even if currentPage > totalPages. Now it returns the next page only if more pages exist. This improves the accuracy of HAL pagination and prevents broken next links.


jsprds (approved):
LGM!

PR #437, opened 2025-06-02
upgrade(ngx-store): Upgrade to latest ngrx-store version

IbenTesara:
**Description**
Upgrade to latest store version

**Requirements**


libs/angular/store/package.json
```diff
@@ -25,9 +25,9 @@
 	"peerDependencies": {
 		"@angular/common": "^19.0.0",
 		"@angular/core": "^19.0.0",
-		"@ngrx/entity": "18.0.0",
-		"@ngrx/store": "18.0.0",
-		"@ngrx/effects": "18.0.0",
+		"@ngrx/entity": "^19.2.1",
+		"@ngrx/store": "^19.2.1",
+		"@ngrx/effects": "^19.2.1",
```
WHeirstrate, line 30: Why is this version not fixed unlike the one in the root `package.json`? Won't it always take 19.2.1 due to the version in the deps having a higher priority than the peerDeps? I'll keep looking for some sources...

IbenTesara: Pretty sure we've been doing it like this for a while now in the other packages, simply to prevent people from getting stuck the way you are currently.

It's a bit of a trade-off and not a perfect solution, but the one that causes the least friction for the users of the package.

PR #442, opened 2025-07-23
feat(ngx-table): add first and last row class to allow for custom styling

WHeirstrate:
**Description**
It is now possible to apply custom styling to the fist and last row of the table, including the detail rows. This would allow for rounded corners, for example. This is showcased in the `After` video recordings below.

This PR also fixes a bug where the `aria-expanded` attribute was still toggled on click even though the global config did not allow for an actual detail row toggle.

**Requirements**

**Attachments**
### Before:

https://github.com/user-attachments/assets/3019d7d8-61cf-492c-bd6b-579cef14184e

### After:

https://github.com/user-attachments/assets/a9914d25-77b2-44df-84d5-b3bf83ca7661


IbenTesara (approved):
Looks good to me, but maybe you should also update the list of classes in the documentation page. 👍

PR #446, opened 2025-07-30
feat(ngx-table): add an optional context row to the table

WHeirstrate:
**Description**
The table has gotten an upgrade!

We required a context row for the VLAIO project and saw no other practical way than to add it as a feature to the table.
In short: a context row is a row that adds no accessibility addition, will always be interpreted as if it was a `span` element. The text in it will be read, buttons and links will have their persisting accessibility features, and an additional row is added to the table. This row allows for a more customisable experience, like the one in the styled example of the docs. (added as PDF to the attachments)

It uses an input and is therefore toggleable. It uses keys defined in the data object array to evaluate whether a row requires an additional context header.

**Requirements**

**PDF of the docs:** 
[context-row-docs.pdf](https://github.com/user-attachments/files/21513717/context-row-docs.pdf)

**Attachments**

https://github.com/user-attachments/assets/01910784-d42c-48da-aa4e-76e6e879cf7d

### DIFF
--- a/libs/angular/table/src/lib/table/ngx-table.component.html
+++ b/libs/angular/table/src/lib/table/ngx-table.component.html
@@ -10,6 +10,9 @@
 	[attr.aria-multiselectable]="selectableType === 'checkbox'"
 	[dataSource]="!loading ? data : []"
 >
+	<!-- ---------------------- -->
+	<!-- TABLE CELL DEFINITIONS -->
+	<!-- ---------------------- -->
 	@if (selectable) {
 		<ng-container cdkColumnDef="ngxTableSelectColumn">
 			<th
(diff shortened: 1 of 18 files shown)


Redn4s (approved):
Ik mis wel wat context, maar code-wise ziet het er wel goed uit. 😅

libs/angular/table/src/lib/table/ngx-table.component.html
```diff
@@ -209,6 +212,22 @@
 			</td>
 		</ng-container>
 	}
+	@if (contextRowTemplate) {
+		<ng-container cdkColumnDef="ngxTableContextColumn">
+			<td
+				*cdkCellDef="let row; let index = dataIndex"
```
reunefe, line 218: Further below you renamed the properties mapping to `dataIndex` to be consistent. Beter do it here as well then

WHeirstrate: Yes, great catch!

libs/angular/table/src/lib/table/ngx-table.component.html
```diff
@@ -251,9 +294,13 @@
 				'ngx-table-row',
 				rowClass || '',
 				row[highlightKey] ? 'ngx-table-row-highlight' : '',
-				first ? 'ngx-table-row-first' : '',
+				first
+					? 'ngx-table-row-first'
+					: !contextRowClasses[dataIndex] && dataIndex === 0
+						? 'ngx-table-row-first'
+						: '',
```
ian-emsens-shd, line 301: These inline if-statements (also in other parts of the code) are getting out of hand. Consider opening an extra issue to split them to utils / consts / ... to preserve legibility

ian-emsens-shd (commented):
In general, I think both this context & detail feature have too much overlap. Consider splitting the way you render your row to a separate component and allowing implementations to overwrite / extend the specifics.

This obviously exceeds the scope of this change and warrants a breaking change but does provide more flexiblit. From what I understand, at the time of writing, the only difference between context & detail is top & bottom respectively.

